Issue #13, opened 2026-06-05
Llamastash doesn't load mmproj files


4c-ee:
## Summary

When calling a model with image support, I noticed it gave the following error:
`Error: API Error (500): image input is not supported - hint: if this is unexpected, you may need to provide the mmproj`
An 'mmproj' is a multimodal projector for the LLM, it's what gives the text model the ability to see and hear through image and audio input.

## Steps to reproduce

1. Find and download a GGUF (in this instance, gemma-4-E4B-it) and an accompanying mmproj (mmproj-gemma-4-E4B-it)
2. Load the model using llamastash
3. Call the model via the server with an image_input.

## Expected behaviour

Llamastash would find the mmproj gguf that accompanies my main model gguf (they're in the same folder), and load it with `llama-server -m /path/to/gemma-4-E4B-it.gguf --mmproj /path/to/mmproj-gemma-4-E4B-it.gguf`.
Like shown in https://github.com/ggml-org/llama.cpp/blob/master/docs/multimodal.md

## Actual behaviour

I had an agent run through the source code real quick and it realized that it's genuinely just not implemented.
> No — llamastash does not currently pass --mmproj to llama-server. This is the gap causing your error.
Here's the full picture:
>1. Discovery filtering (src/discovery/scanner.rs:124-129): mmproj-*.gguf files are detected by filename prefix and excluded from the model catalog so they don't appear as launchable rows.
>2. Launch composition (src/launch/params.rs:443-493): The compose() function builds the argv for llama-server. It passes -m <model_path>, --host 127.0.0.1, --port, --embeddings/--reranking per mode, --jinja --reasoning-format, -c, then typed knobs and extras. There is no --mmproj argument anywhere.
>3. LaunchParams struct (src/launch/params.rs:131-169): Has model_path, mode, ctx, reasoning, knobs, and extras. No mmproj_path field.
>
>So the flow is: you select model.gguf → llamastash launches llama-server -m model.gguf ... → llama-server has no vision projector and says "image input is not supported — hint: you may need to provide the mmproj."
>
>What's needed to fix this:
>1. A function to find the companion mmproj-*.gguf next to the model (by same directory, matching prefix pattern)
>2. An mmproj: Option<PathBuf> field in LaunchParams
>3. In compose(), emit --mmproj <path> when mmproj is Some
>4. Thread it through from the TUI/CLI start path into ManagedSpawn → LaunchParams

## Environment

- `llamastash --version`:
llamastash 0.0.2
[...]

PR #15, opened 2026-06-06
feat: find and load mmproj ggufs that are stored alongside models

4c-ee:
Resolves issue #13

## What
Auto-detects multimodal projector (mmproj) companion files when launching a model.

## Why
Multimodal models (Gemma 4, Qwen 3.6, etc.) ship the projector 
as a separate `.gguf` file alongside the main model. In llama.cpp, users 
pass `--mmproj` manually, but now llamastash will do it for them.

## How
- `scanner::find_mmproj()` scans the model's parent directory for `mmproj-*` 
  or `mmproj_*` companions
- `start_model` IPC handler uses this as a fallback when `mmproj_path` is 
  not explicitly provided
- `--mmproj <path>` is appended to the llama-server argv

## Testing
- `cargo test --features test-fixtures` — all pass
- Manual: launched a Qwen2.5-VL model with its `mmproj-Qwen2.5-7B.Q4_K_M.gguf` 
  companion in the same directory
- Built release and tested on Arch Linux, multimodal input works great

Forgive any mistakes I might've made, still new to git & github lol

### DIFF
--- a/src/discovery/scanner.rs
+++ b/src/discovery/scanner.rs
@@ -8,10 +8,11 @@
 //! starve the runtime when the user has hundreds of GGUFs on disk.
 
 use std::path::{Path, PathBuf};
-use std::sync::Arc;
+use std::sync::{Arc, OnceLock};
 
 use futures::stream::{self, StreamExt};
 use ignore::WalkBuilder;
+use regex::Regex;
 use tokio::sync::mpsc;
 
 use crate::discovery::metadata_cache::{self, CachedParse, MetadataCache};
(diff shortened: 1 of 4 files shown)


damiensawyer:
Hi @4c-ee  :-)

I hadn't heard of mmproj files. I guess that's why my local models have been struggling with vision!! 

I asked Claude to summarize this PR, as in explain what it did. 

In addition, and unprompted, It included these suggestions. I don't know if they're valid or not. I've very new to rust. 

---------------

Things worth flagging if you're reviewing

  - The name-matching is very strict. It requires the mmproj filename's remainder to exactly equal the model stem (e.g. model model.gguf →
  mmproj-model.gguf). In the real world mmproj files usually don't mirror the model stem — Bartowski/unsloth repos ship things like
  mmproj-model-f16.gguf or mmproj-Qwen2-VL-7B-Instruct-f16.gguf next to a ...-Q4_K_M.gguf model. Those would be missed, so auto-detection may
  silently not fire on most real downloads. The PR's own example (mmproj-Qwen2.5-7B.Q4_K_M.gguf beside Qwen2.5-7B.Q4_K_M.gguf) is the lucky case
  where stems match.
  - It duplicates existing logic. The scanner already has is_projector_companion() and a test collect_gguf_paths_drops_mmproj_projector_companions
  — i.e. the codebase already recognizes mmproj files (to exclude them from the catalog). The new function doesn't reuse that detection.
  - Minor robustness: find_mmproj uses entry.ok()? and name.to_str()? inside the loop — a single non-UTF8 filename or unreadable entry will abort
  the whole scan and return None rather than skipping that one entry. continue-on-error would be more resilient.
  - The or_else correctly makes explicit mmproj_path win over auto-detect, which is the right precedence.

  Net: small, well-tested, correctly plumbed mechanism. The main open question is whether the strict stem-matching is loose enough to actually
  trigger on typical HuggingFace multimodal repos — that's the thing I'd push back on before merging.

4c-ee:
Hi! Claude is actually right, `mmproj-/_*` is too strict. I'll try to find the most common types of filenames on huggingface, and come up with some far more robust file matching.

deepu105:
Thanks @4c-ee and @damiensawyer

deepu105:
i'm gonna push a commit to fix the CI fails and add some tests and fix some edge cases

deepu105:
This is a great addition to the project @4c-ee 
Thank you :pray:

deepu105 (approved):

PR #49, opened 2026-06-27
fix(daemon): inherit extras from last_params on proxy auto-start

Silvani-In-The-Net:
When a model is unloaded and reloaded by the proxy auto-start path (e.g. agent switches), free-form argv flags like --chat-template-file were lost because they are persisted in last_params but never read back.

The fix adds an inheritance path for extras similar to how last_params_knobs is already read: if parsed.extras is empty (proxy auto-start sends an empty vec), fall back to the model's last_params extras so free-form flags survive reload cycles.

Build was tested and confirmed to work as expected, there might have been an edge case I'm not smart enough to think of like manual unload discarding the extras upon proxy reload.

### DIFF
--- a/src/daemon/launch_service.rs
+++ b/src/daemon/launch_service.rs
@@ -26,7 +26,7 @@ use crate::daemon::registry::LaunchId;
 use crate::daemon::shutdown::ShutdownToken;
 use crate::daemon::state_store::RunningSnapshot;
 use crate::daemon::supervisor::{
-  spawn as supervisor_spawn, ManagedModel, ManagedSpawn, ManagedState,
+  spawn as supervisor_spawn, LaunchOrigin, ManagedModel, ManagedSpawn, ManagedState,
 };
 use crate::gguf::header::{read_path as read_gguf_header, HeaderReadOptions};
 use crate::gguf::identity::ModelId;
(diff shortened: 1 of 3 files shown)


src/daemon/launch_service.rs
```diff
@@ -258,7 +258,21 @@ pub(crate) async fn compose_and_spawn(
   // (identity rule); an explicit choice from `start --backend` / the TUI
   // picker overrides it. Resolved into a backend at the selection seam below.
   launch_params.backend = parsed.backend.unwrap_or_default();
-  launch_params.extras = parsed.extras.into_iter().map(OsString::from).collect();
+  // Resolve extras from the caller's request. If the caller didn't provide
+  // extras (e.g. proxy auto-start), inherit from the model's last_params so
+  // free-form flags (`--chat-template-file`, `--mmproj`, etc.) survive reload
+  // cycles. When the caller *does* provide extras — including an empty vec —
+  // use them verbatim (the TUI always sends extras, even when cleared).
+  launch_params.extras = if parsed.extras.is_empty() {
```
deepu105: The empty-vs-set check works for the proxy, but that signal isnt unique to it. A normal `start_model` (a TUI launch or `llamastash start <model>`) with no passthrough flags also sends an empty vec, so it'll now re-inherit whatever extras the last launch used, and there's no way left to start with zero extras once last_params holds some.

`compose_and_spawn` already takes an `origin: LaunchOrigin` for this exact split: the proxy auto-start path passes `AutoStart` and the IPC `start_model_handler` passes `Manual`. If we gate the inheritance on `origin` being the auto-start case instead of `is_empty()`, the fix stays scoped to the reload path and manual launches keep using their extras verbatim. Wdyt?

Side note: the comment just above (lines 264-265) says an empty vec is used verbatim, but the code inherits on empty, so that one will want a tweak too whichever way we land.

src/daemon/launch_service.rs
```diff
@@ -258,7 +258,21 @@ pub(crate) async fn compose_and_spawn(
...
+    let snap = ctx.state.snapshot().await;
```
deepu105, line 264: Minor, not a blocker: `snapshot()` clones the whole `DaemonState` under the lock, and there's a second `ctx.state.snapshot().await` a few lines down (the last_params_knobs block, around line 292) that reads the same `last_params_map().get(&identity)`. Since we're already here, we could grab both extras and knobs from a single snapshot instead of cloning the full state twice per launch.

deepu105 (commented):
LLM review: thanks for chasing this down, the dropped extras on proxy reload is a real bug and falling back to last_params is the right instinct. I ran an LLM pass over the diff and there's one thing I'd want to sort before merging, plus a small cleanup. Left both inline.

Short version: branching on `parsed.extras.is_empty()` cant tell a proxy auto-start (always empty) apart from a normal launch that genuinely has no extras, so it quietly changes behavior for manual launches too. Good news is the function already gets a signal that draws exactly that line.

deepu105:
Hi @Silvani-In-The-Net Thanks for the PR. Please see the review comments posted by my harness. This also made me find an existing issue were preset extras are also dropped. i'll fix that after merging this one. You mentioned you are new to rust open source, let me know if you would rather like me to make the fixes or if you want to iterate on it. Happy to help eitherway :pray:

Silvani-In-The-Net:
Yes, I am quite new to rust open source, until I get my bearings I'll divert the changes to you instead of taking a crack myself.

Thank you for the critiques!!

deepu105:
@Silvani-In-The-Net no worries I can make the changes and merge. Thanks again for fixing this issue. Appreciate it :pray:

deepu105 (approved):

PR #55, opened 2026-07-29
feat(gpu,daemon): add GPU probe tuning, idle shutdown, and configurable sample interval

qo0u0op:
## What

Four new config-only keys (no CLI/env surface) for GPU probe and daemon lifecycle tuning:

| Key | Default | Description |
|-----|---------|-------------|
| `gpu.enable_vulkan_probe` | `true` | Skip Vulkan fallback probe (`vulkaninfo`) when `false` |
| `gpu.reprobe_interval_secs` | `60` | Full vendor re-probe interval; `0` disables periodic re-probes |
| `daemon.idle_timeout_secs` | `0` | Auto-shutdown after N idle seconds; `0` = never |
| `daemon.probe_interval_secs` | `1` | Host-metrics tick interval (1–60s) |

## Why

- **Vulkan probe skip** — redundant when a native NVIDIA/AMD/Metal probe covers all GPUs; avoids unnecessary `vulkaninfo` subprocess spawns.
- **Reprobe interval** — personal desktops don't need per-minute hotplug detection; raising this reduces `nvidia-smi`/`rocm-smi` calls.
- **Idle shutdown** — forgets `daemon stop` on a laptop wastes battery; opt-in auto-shutdown after configurable inactivity.
- **Sample interval** — lowers dashboard refresh frequency on low-power machines at the cost of less responsive numbers.

## Notes

- All keys are config-only, documented in `config.example.yaml`, `docs/usage.md`, and `docs/troubleshooting.md`.
- `probe_interval_secs: 0` is clamped to `1` at both the CLI layer and `host_metrics::spawn` to prevent a CPU-burning busy loop.
- `reprobe_interval_secs` skips periodic re-probes on `CpuOnly` and `Unknown` (Vulkan-only) backends — they have no live metrics to refresh.


deepu105:
Thank you for the nice additions. Will check and merge

PR #60, opened 2026-08-07
fix(gpu): keep coherent-UMA NVIDIA devices instead of dropping them

### LINKED ISSUE #59: Nvidia DGX Spark GPU not recognized
I am running llamastash 0.0.6 on an MSI DGX Spark with a Grace/Blackwell processor. When I run llamastash init i expected the wizard to identify the GPU and understand the RAM is available a VRAM - it does not, it only sees the CPU. It also did not find the llama.cpp install on the device, I had to use the command line parameter --llama-server with the fully qualified path for it to work. Also note that the LLAMASTASH_PROXY_HOST command line arg shows unknown when passing, had to set that as an ENV variable and pass to the runtime that way for it to bind on 0.0.0.0.

Image

bernardogv:
Fixes #59.

On GB10 / Grace Blackwell (DGX Spark) and Jetson, CPU and GPU share one physical memory pool, so `nvidia-smi` reports the memory columns as `[N/A]` — there is no framebuffer to size:

```
$ nvidia-smi --query-gpu=name,memory.total,memory.used,utilization.gpu,temperature.gpu --format=csv,noheader,nounits
NVIDIA GB10, [N/A], [N/A], 0, 39
```

`parse` treated the unparseable size as a malformed row and `continue`d, discarding the whole device. `probe_devices` then saw an empty vec and returned `None`, so no NVIDIA GPU was detected at all — `doctor` printed `GPU  CPU only` on a machine with a GB10 in it, and reported 0 findings while doing so.

The asymmetry looked unintentional: `used`, `utilization` and `temperature` all tolerate the markers, and the module doc already said the parser stores `None` for them rather than skipping the row. Only `memory.total` hard-failed.

### What changed

- Only the **bracketed markers** route to the UMA path. Genuine garbage in the column still drops the row — pinned by a test, so this doesn't quietly promote malformed input to a device.
- Memory for those devices is read from the system pool, which is what they allocate from. **Available** rather than free: the page cache is reclaimable and the gap between the two routinely runs to tens of GiB on these parts, so `free` would badly understate what a model can use.
- The pool is also recorded as `uma_shared_*`, so the host pane doesn't count the same bytes twice against the RAM gauge.
- The verdict gets its own `ClassSource::NvmlNoFramebuffer` rather than reusing the AMD carve-out heuristic. NVML refusing to report a framebuffer is an *authoritative* signal — a discrete card always reports a size — so it sits with `AppleUnified` / `ExplicitDxgiUma`, not with the inferred `CarveSignature`.

### Verified on hardware

DGX Spark / GB10, driver 580.173.02, CUDA 13.0, aarch64, Ubuntu 24.04.4, kernel 6.17.0-1029-nvidia:

```
BEFORE                              AFTER
  CPU   Cortex-A725 · 20 cores        CPU   Cortex-A725 · 20 cores
  MEM   121.7 GiB                     MEM*  121.7 GiB
  GPU   CPU only                      GPU   NVIDIA · 121.7 GiB (unified)
  DISK  1320.1 GiB free               VRAM (shared) 121.7 GiB
                                      DISK  1318.4 GiB free
```

```json
// doctor --json, hardware section
{ "gpu_backend": "nvidia", "unified": true,
  "uma_class_source": "nvml_no_framebuffer",
  "gpu_pool_total_bytes": 130663165952, "mem_total_bytes": 130663165952 }
```

The `MEM*` marker appears without any change to display code — setting `uma_shared_*` is enough for the existing host-pane logic to treat the pool as shared.

### Checks

- `cargo test --features test-fixtures` — 2053 lib + all integration suites pass
- `cargo clippy --all-targets --features test-fixtures -- -D warnings` — clean
- `cargo fmt --all -- --check`, `make doc` — clean
- 5 new unit tests in `gpu::nvidia::tests` covering: the GB10 row is kept and classified, discrete cards stay unclassified, non-marker garbage still drops, the fill populates from the pool, and the fill leaves discrete cards alone

### Notes

- Per CONTRIBUTING I commented on #59 before opening this. Happy to close it if you'd rather take the fix a different way — the diagnosis is in the issue either way.
- Docs updated in the same change per AGENTS.md: `CHANGELOG.md`, plus two statements this falsifies — `GpuDevice.classification_source`'s doc said NVIDIA never classifies, and `AGENTS.md`'s `uma_class_source` value list omitted the new variant.
- I wasn't sure which Implementation Unit `src/gpu/` belongs to, so the commit is scoped `fix(gpu):` — happy to relabel.
- Unrelated, spotted while working: `docs/troubleshooting.md` "GPU not detected" tells NVIDIA users to install NVML, but this path shells out to `nvidia-smi` (the NVML follow-up is still a TODO in `gpu/mod.rs`). Left it alone to keep this diff focused, but it would have sent the reporter down the wrong path.
- I can test further revisions on real GB10 hardware — just push to the branch or point me at one.


deepu105:
Thank you @bernardogv appreciate it

deepu105:
i'll merge the fix to main and include it in  next release (hopefully today), appreciate any more testing on NVidia hardware as I'm on AMD stack

deepu105 (approved):
**Kimi K3**: Reviewed the diff, traced every consumer of the touched fields, and ran the code. **LGTM with two nits — approving.**

## Verification

- Applied the patch to a local tree (one trivial CHANGELOG conflict with `d462e82`, both `### Fixed` lines kept).
- `cargo test --features test-fixtures --lib` — 2060 passed, including the 5 new `gpu::nvidia` tests.
- `cargo clippy --all-targets --features test-fixtures -- -D warnings` and `cargo fmt --check` — clean.
- E2E on an AMD Strix Halo host (no NVIDIA here, so the GB10 path itself rests on your on-hardware verification): `doctor` + `status --json` still report `carve_signature` / `unified` / GTT pool correctly — no regression to the existing UMA path.

## What I checked beyond the diff

- **Admission control** (`launch/admission.rs::effective_free_bytes`): on a GB10 the unified branch computes `min(ram_free, gtt_free)`. On Linux, sysinfo 0.39's `used_memory = total − available`, so both legs resolve to `MemAvailable` — the "available, not free" choice survives the whole pipeline, not just the probe. Consistent with the `IntegratedUma` 1.0 usable-fraction policy (budgeting *free* bytes, not total).
- **TUI VRAM gauge** (`vram_denominator`): `gpu_in_shared = uma_shared_used` exactly cancels `ram_used`, so the denominator is the full pool and the gauge reads system-used / pool — the honest rendering of one coherent pool. Consequence worth knowing: on GB10 the VRAM row mirrors the MEM row. I think that's right for a shared pool, but it is a semantic difference from AMD (GPU-allocated only) and Apple (no used figure).
- **Version skew**: `cli/output.rs` deserializes `uma_class_source` with `.ok()`, so an older client reading `nvml_no_framebuffer` from a newer daemon degrades to `None` instead of failing. No compat concern.
- **Per-tick cost**: `fill_unified_memory` runs inside `probe_devices`, which the 1 Hz sampler hits via `refresh_active` — a RAM-only sysinfo refresh is a `/proc/meminfo` read, negligible.
- **Dedup**: GB10/Jetson are SoC-only parts, so the "every UMA device gets the full pool" sum can't double-count on real hardware; Windows DXGI overlap is handled by the existing PCI/name dedup.

## Nits (non-blocking)

1. **`is_unsupported_marker` is broad** — any `[`-prefixed value in `memory.total` classifies as UMA. The two markers nvidia-smi actually emits here are `[N/A]` and `[Not Supported]`; a hypothetical `[Insufficient Permissions]` on a discrete card would be promoted to a UMA verdict and get its "VRAM" filled from system RAM. Matching the two known markers exactly would be tighter. Low likelihood, and your garbage-drop test pins the other direction — your call.
2. **Stale comment** in `gpu/mod.rs::uma_class_source` (line ~356): "e.g. NVIDIA, which doesn't classify" — this PR is exactly what makes that imprecise. One-line fix while you're here.
3. **CHANGELOG** entry is 4 lines of implementation detail; AGENTS.md asks for short one-liners. Something like "NVIDIA coherent-UMA GPUs (GB10 / DGX Spark, Jetson) are no longer dropped from detection (#59) — `nvidia-smi` reports no framebuffer on these shared-pool parts and the probe used to discard the whole device" keeps the signal.

The `docs/troubleshooting.md` NVML-vs-nvidia-smi mismatch you flagged is real but pre-existing — fine to leave out of this diff; worth a TODO.md entry per the repo rule so it doesn't get lost.

Nice work — the `ClassSource::NvmlNoFramebuffer` variant (authoritative signal, not the carve heuristic) is the right call, and the test coverage pins both directions.

bernardogv:
**v0.1.0 verified on DGX Spark (GB10) hardware.** Built the release from `c5e3b76` in a clean container and ran `doctor` with `--gpus all` — same procedure as the original repro on #59:

```
hardware
  CPU   Cortex-A725 · 20 cores · NEON SVE
  MEM*  121.7 GiB
  GPU   NVIDIA · 121.7 GiB (unified)
  VRAM (shared) 121.7 GiB
  DISK  886.7 GiB free
  OS    linux/arm64
llamastash doctor (0 findings)
✓ everything looks healthy
```

Previously (`v0.0.6`, same box): `GPU   CPU only`. The device is now detected and correctly classified as coherent unified memory with the full 121.7 GiB reported. Happy to test future NVIDIA-side changes on this hardware — feel free to ping me.

deepu105:
Thank you @bernardogv appreciate it

PR #68, opened 2026-08-24
fix: launch llama.cpp's unified `llama` binary

### LINKED ISSUE #67: Llama.cpp 0.2.0 release
With the new release of Llama.cpp which lets you install the app on Windows using irm or Winget, llama.exe is a combined binary with subcommands. So the command has changed to:

`> llama.exe serve -m "model.gguf"`

The binary llama-server.exe is still available separately, but I would like to use the new install method which automates updates.

Is there a way to run this new command? Thank you. I am a beginner.

deepu105:
Closes #67.

llama.cpp's own installer at [llama.app](https://llama.app) now ships a single `llama` / `llama.exe` that serves behind a `serve` subcommand, and it installs no `llama-server` at all. That is what the `irm` path on Windows leaves you with, and `install.sh` does the same thing on Linux and macOS, dropping `llama` in `~/.local/bin`. llamastash only ever looked for `llama-server`, so a fresh install resolved nothing and `start` exited 70.

This is not Windows-only. The unified binary is in the current release tarballs for macOS, Linux and Windows (checked b10604, b10605 and b10647), so the fix is one name-based dispatch with no OS-gated path.

## What changed

`Backend::process_marker` becomes `process_markers`, a list of the basenames one backend's server can carry, primary first. llama.cpp declares `["llama-server", "llama"]`.

- The `$PATH` locator walks the list in order, so an install carrying both binaries still resolves `llama-server` and nothing about an existing setup moves.
- The orphan sweep learns the second name for free, so a hand-started `llama serve` is recognised as an external process instead of being invisible.
- `--llama-server`, `LLAMASTASH_LLAMA_SERVER` and `backend.llamacpp.servers[]` accept either binary.
- `init`'s common-location probe checks both names in each directory it already searched.

`llama_cpp::serve_prefix` adds the subcommand ahead of the flags, for the launch argv and for the `--list-devices` probe (`llama --list-devices` is an unknown command, `llama serve --list-devices` is not).

Dispatch is on the binary's name rather than its version, because both binaries ship in the same release and report the same version string, so a version check cannot tell them apart. The flags are identical either way: `llama serve --help` is byte-identical to `llama-server --help`, so nothing else in the tree branches on which one resolved.

## Verified

Against a local ROCm build 10610 (`a14dba686`), with only `llama` on `$PATH` and no `llama-server` anywhere:

```
binary   /home/deepu/.cache/llamastash-e2e-67/unifiedbin/llama
server   llamacpp-rocm, ROCm0: AMD Radeon 8060S Graphics (126976 MiB)
argv     llama serve --host 127.0.0.1 --port 41100 -m ...Llama-3.2-1B-Instruct-Q4_K_M.gguf --jinja --fit-ctx 16384
```

The model reached ready in 4.4s, answered on the proxy, and a hand-started `llama serve` on another port showed up in `status --external`. Putting `llama-server` back on `$PATH` produces the argv it always did, with no `serve` token. Also checked the `LLAMASTASH_LLAMA_SERVER` route pointed at the unified binary, and the TUI panes.

`make test` passes.

## Review follow-up

`fix(unit5): only count the unified binary's serve mode as a server` picks up the review:

- **The winget claim was wrong and is gone** from the CHANGELOG, `INSTALL.md`, `docs/troubleshooting.md`, `docs/architecture.md` and the `SERVER_BINARIES` comment. The only winget package is `ggml.llamacpp`, whose manifest installs `llama-server.exe` plus tools, so a winget install never hits this bug. The llama.app installer is the one that ships only `llama`.
- **The external sweep no longer claims the unified binary's other modes.** Matching a swept process on its basename alone reported a live `llama cli` or `llama download` session as an unmanaged llama.cpp server, and made it a legal explicit `stop` target. A new `Backend::argv_is_server` hook re-checks the argv for the subcommand the backend would have launched with. The standalone binary keeps matching on basename alone, so nothing changes for it, and a process whose argv the OS will not hand over keeps the basename verdict.
- **The not-found error stopped hardcoding binary names.** It reads them back from the same marker list the `$PATH` search walks, so it names no binary itself.
- `docs/architecture.md` pointed at `src/backend/llama_cpp.rs`; the module is a directory.

Checked with two live processes of the same basename, upstream b10605 on an isolated daemon:

```
llama serve  -m ...Llama-3.2-1B-Instruct-Q4_K_M.gguf --port 41234   listed, port parsed
llama cli    -m ...Llama-3.2-1B-Instruct-Q4_K_M.gguf                not listed
llama-server --port 41100 ...                                       listed, port parsed
```

Reverting just the argv gate and rebuilding puts the `llama cli` row back, so the case is real rather than theoretical.

### Second round

`refactor(unit5): keep binary-name knowledge inside the backend registry` picks up the three nits:

- **`init` no longer reaches into a backend module.** The adopt-path hint imported `llama_cpp::serve_prefix` to decide whether a hand-picked path could serve, which was the one code reference to a backend module outside the sanctioned three. It now asks `Backend::binary_serves`, whose default answers from `process_markers` plus the existing "contains server" tolerance for custom builds. `basename_matches_marker` moved from the orphan sweep to the registry so the sweep and the hint share one rule, rather than two copies of the short-marker hazard.
- **A re-adopted row prints the real command line.** It was reconstructed from the primary marker, so a unified-binary launch read as `llama-server --port N` when the process was actually `llama serve --host ...`. The pid lookup already pulled full process metadata for the start time, so the argv came for free; the reconstruction stays as the fallback for a process the OS no longer has.
- The third note needed no change.

Same before-and-after check, on one adopted process (upstream b10665, isolated daemon, `kill -9` on the daemon to orphan the child, then restart so the boot sweep re-adopts it):

```
before   llama-server --port 41101 -m ...Llama-3.2-1B-Instruct-Q4_K_M.gguf
after    /home/.../llama-b10665/llama serve --host 127.0.0.1 --port 41101 -m ... --jinja -c 4096
```

`binary_serves` keeps every case the old hint got right, `llama-cli` still warns and the unified `llama` still does not.

## Also in this PR

Two commits for CI failures that are red on `main` right now, unrelated to the fix above:

- `chore(deps): h2 0.4.18 to clear RUSTSEC-2026-0258`. The advisory landed on 2026-08-17 and fails `cargo audit --deny warnings`. h2 is transitive via hyper and reqwest, nothing pins it, lockfile only.
- `test(init): stop the symlink test asserting on a real /tmp`. `a_symlinked_config_is_written_through_not_replaced` used `env::temp_dir()` as a default path it expected back unchanged, but on the macOS runner that is `/tmp`, which is a link to `/private/tmp`. `resolve_symlinks` was doing the right thing and the test was asserting the wrong premise, so both macOS test jobs fail. The dummy patcher now defaults to a path inside the test's own temp dir.


deepu105 (commented):
Ran an independent pass on this, from scratch, against real artifacts rather than the PR body. Everything held up except one claim.

**Verified against the real thing** (local ROCm build 10610 `a14dba686`, upstream at b10647 when checked):

- `llama serve --help` vs `llama-server --help`: byte-identical, 703 lines
- both binaries report the same version string, so name-based dispatch is the only option
- `llama --list-devices` fails, `llama serve --list-devices` works
- E2E on an isolated state dir with only `llama` on PATH: resolved, device probe enumerated the 8060S, spawned argv was `llama serve --host 127.0.0.1 --port N -m ...`, model reached ready, proxy answered a real `/v1/embeddings` round-trip
- both binaries on PATH resolves `llama-server`; an only-`llama-server` PATH produces the old argv with no `serve` token; the `--llama-server` flag canonicalizes a symlink to the target and prepends `serve`
- hand-started `llama serve` on :41234 showed up in the external sweep with port and model parsed
- `install.sh` and `install.ps1` from llama.app install only `llama` (`~/.local/bin/llama`, `%LOCALAPPDATA%\llama-app\llama.exe`)
- RUSTSEC-2026-0258 matches the commit message exactly; h2 was 0.4.14 on main, 0.4.18 here
- main CI run 32723570497 fails exactly the three named jobs: `cargo audit` plus both macOS test jobs
- `make lint` and `make test` pass locally

**One claim is wrong: winget.** The only winget package is `ggml.llamacpp`, and its newest manifest (b10615, released the same day as this PR) installs `llama-server.exe` plus tools, not `llama.exe`. A winget install still gets `llama-server` and never hits this bug. The `irm` / llama.app part is right, and the tarballs do carry both binaries (checked b10604, b10605, b10647 ubuntu, b10647 macos-arm64 and win-cpu-x64; b10604 already has `llama`, so the b10605+ boundary in the body is loose). The winget mention should come out of the CHANGELOG, INSTALL.md, troubleshooting.md, architecture.md, and the `SERVER_BINARIES` comment.

**One behavior worth a conscious decision:** the external sweep matches on executable basename only, so a `llama cli` or `llama download` session now surfaces as an external llama.cpp server row and is a legal explicit `stop` target. Impact is contained, `stop --all` only touches managed launches and externals need an explicit pid, and `llama-cli` / `ollama` are correctly excluded by the `marker.contains('-')` guard. The sweep already reads `proc.cmd()`, so a `cmd[1] == "serve"` filter was available. Fine to ship as is, but that is a choice, not an accident.

Nits: `locate`'s NotFound error now hardcodes both binary names in `src/launch/binary.rs`, which stretches the no-leak rule even though the old message already named one. And the architecture.md line still points at `src/backend/llama_cpp.rs`; the module is a directory.

PR #71, opened 2026-08-25
feat(proxy): add proxy.max_body_size and raise default cap to 16 MiB

### LINKED ISSUE #65: Allow configuring Proxy Body Size Limit
## Problem

When running a vision-enabled model via the proxy (in my case Qwen 3.8 27B for use in opencode), I hit the proxy body size limit whenever I try and attach an image that is 2+ MiB in size. According to the docs at https://github.com/llamastash/llamastash/blob/main/docs/usage.md#endpoints, this limit was intentionally chosen since "OpenAI chat completion requests are typically well under 1 MiB even with long histories", which might be true for text only cases, however in my experience it is trivially easy to accidentally run into the limit once you start attaching images.

## Proposed solution

Allow setting the proxy body size limit in the configuration file. An example configuration key could be `proxy.max_body_size` with a size in bytes. So for example for a 10 MiB limit, I could write:

```yaml
proxy:
  max_body_size: 10485760
```

## Alternatives considered

- **Increase the hard-coded limit:** Would also work for my case but I assume there's a reason the original limit was chosen to be 2 MiB and if you're not using a vision-enabled model, it might not be necessary to increase this.
- **Do nothing, Users manually downsize all images before sending them to stay below the limit:** Tedious to do every time, images might also lose details leading to potentially worse responses
- **Do nothing, just don't send images:** But I want to 😢 

## Scope check

- [x] I checked the requirements + plan docs and this is not already in scope
- [ ] This is v2+ work (HTTP, MCP, HF pull worker, etc.) and that's OK

The limit is mentioned in `2026-05-21-001-feat-proxy-router-plan.md` with the remark "Not exposed in [proxy] config at v1." I think this decision might not have taken vision models into account and their need for larger body sizes in order to transfer images, which are mostly blocked by the 2MiB limit. Therefore I would encourage you not to defer this feature to v2+ if it was ever originally intended to eventually be exposed in the first place.

deepu105:
## Summary

Vision payloads stopped dying at the proxy. The proxy buffered every request body under a hard-coded 2 MiB cap, so a base64 image (10 MiB is typical for a phone photo) got `413 payload_too_large` before ever reaching the upstream. Now the cap is a per-daemon config key and the default is 16 MiB, so vision requests work out of the box.

Closes #65.

## What changed

- New `proxy.max_body_size` key: raw bytes, threaded `ProxyConfig` to `ProxyState` into the three buffered paths (chat forward, Ollama `/api/show`, `/ui`). `0` is legal and documented: it rejects every non-empty body.
- Default raised from 2 MiB to 16 MiB (`DEFAULT_BODY_LIMIT_BYTES`). One base64 phone photo plus long history fits; two concurrent max-size requests bound worst-case memory at 32 MiB.
- The 413 message now names the configured cap (for example `request body exceeds the 2 MiB limit`) instead of a fixed unit, so an operator can tell which cap fired.
- Deferred per the issue: no CLI or env override, no `status` field, no per-route caps. Config stays raw bytes; the human-readable unit appears only in the error message.

## Upstream wall (recorded, not fixed here)

`llama-server` itself caps request payloads at 1 MiB (compile-time `CPPHTTPLIB_FORM_URL_ENCODED_PAYLOAD_MAX_LENGTH`, still present on upstream master). That wall is pre-existing and separate from the proxy cap: the proxy now delivers the full body to it, but llama.cpp-routed bodies over 1 MiB still 413 upstream, with an empty body and a `server: llama.cpp` header. `lemond` does not share the wall (verified with a 3.2 MiB body). The full boundary probe, including which binary versions were tested, is in the E2E record.

## Test plan

- 2235 tests green: loader default and roundtrip, `format_bytes` unit rendering, 413 with a 1 KiB cap naming that cap, a 3 MiB body passing the old default (now 503 `launch_failed`, not 413), and the example config still parsing.
- Real-hardware E2E (isolated daemons, real `llama-server` and `lemond`, no fakes): a 10 MiB body traverses the default-cap daemon to the upstream byte-for-byte; a 2 MiB-cap daemon returns 413 naming the cap; a 0.9 MiB body auto-starts `llama-server` and completes. Record: `docs/testing/2026-08-24-e2e-issue65-proxy-max-body-size.md`.

Plan: `docs/plans/2026-08-24-001-feat-proxy-max-body-size-plan.md`

---

Authored by Qwen3.8-27B-ROCmFP4-FAST, running locally through llamastash.


src/proxy/route.rs
```diff
@@ -193,24 +198,47 @@ pub(crate) async fn buffer_body(body: Incoming, cap: usize) -> Result<Bytes, Bod
...
       super::router::error_response(StatusCode::BAD_REQUEST, "invalid_request", &message)
     }
   }
 }
 
+/// Render a byte cap for the 413 message: largest unit the value
+/// fits in, whole when exact (`16 MiB`), two decimals otherwise with
+/// trailing zeros trimmed (`976.57 KiB`, `2 MiB`, never `2.00 MiB`).
+/// Sub-KiB caps stay raw bytes.
+fn format_bytes(n: usize) -> String {
```
deepu105: `{:.2}` rounds to nearest, so a cap sitting just under a unit boundary renders as the boundary:

```
2097151 bytes (2 MiB - 1)  ->  "2 MiB"
1048575 bytes (1 MiB - 1)  ->  "1024 KiB"   // while 1048576 -> "1 MiB"
```

So a client sending exactly 2097152 bytes against a 2097151 cap gets told "request body exceeds the 2 MiB limit", which reads as a contradiction since their body is 2 MiB. It only bites on non-round caps, and every value in the plan and the E2E record (1 KiB, 512 KiB, 10 MiB, 16 MiB, 30 MiB) renders exactly, so this is low stakes. Still a confusing trail for anyone who picks an odd number.

Cheapest fix is to carry the raw count as well:

```
request body exceeds the 16 MiB (16777216 bytes) limit
```

That reads better for agents parsing the error too. And if you do it, `init::detection::fmt_bytes` becomes good enough and this function can go, which matters a little because this is now the fourth byte formatter in the tree (`init::detection::fmt_bytes`, `tui::fmt::format_bytes`, `backend::vllm::human_bytes`, this one), each with its own output rules. The doc comment on `fmt_bytes` exists precisely because earlier copies drifted on units and decimals.

src/proxy/ui.rs
```diff
@@ -218,12 +218,13 @@ async fn forward_ui(
   entry: RunningEntry,
 ) -> ProxyResponse {
   let (method, uri, headers, body) = forward::deconstruct(req);
-  // Buffer the body under the shared 2 MiB cap via the same helper the
+  // Buffer the body under the daemon's cap via the same helper the
   // data plane uses. UI asset GETs are bodyless; the base-relative API
   // POSTs carry a JSON payload we pass through untouched.
-  let body_bytes = match route::buffer_body(body, route::BODY_LIMIT_BYTES).await {
+  let cap = state.max_body_size;
```
deepu105, line 224: `forward_ui` is the only `buffer_body` caller and it's now the only capped path with zero test coverage.

The plan said to add a cap case to `tests/proxy_ui.rs` if a body-bearing POST existed, and none does (`grep -n POST tests/proxy_ui.rs` returns nothing), so skipping it was a correct read of the plan. The side effect is that if someone later refactors this back onto a constant, nothing fails. One POST against `/ui/v1/chat/completions` with a low cap would close it.

src/config/loader.rs
```diff
@@ -341,6 +341,16 @@ pub struct ProxyConfig {
...
+  /// Cap in bytes on every request body the proxy buffers before
+  /// forwarding (`/v1/*`, `/api/show`, `/ui`). The default 16 MiB
+  /// covers vision payloads (a base64 image is ~33% larger than the
+  /// source file) while still bounding worst-case memory; `0` rejects
+  /// every non-empty body (HTTP 413 on any body). Anything larger
+  /// returns HTTP 413 `payload_too_large` naming this limit.
+  ///
+  /// Sources — CLI: (none) · Env: (none).
+  #[serde(default = "ProxyConfig::default_max_body_size")]
+  pub max_body_size: usize,
```
deepu105, line 353: Two things worth pinning down here, since this is the surface users actually touch.

`0` meaning "reject every non-empty body" collides with how the sibling keys read. `idle_ttl_secs: 0` disables eviction, `--proxy-port 0` picks an ephemeral port. Someone reaching for `max_body_size: 0` expecting "no limit" bricks their proxy instead. It's documented in three places now and locked in by `proxy_config_max_body_size_zero_parses`, so changing it later is a breaking change.

There's also no upper bound. `max_body_size: 99999999999` parses fine and lets one request buffer arbitrary RAM, which ties back to the concurrency point in my top-level comment.

Naming nit while we're here: every other unit-carrying key in this struct is suffixed (`header_read_timeout_secs`, `idle_ttl_secs`), so `max_body_bytes` would match. The issue proposed this spelling and it's already in `config.example.yaml` plus two docs, so I'm not pushing hard, just noting it's inconsistent with its neighbours.

docs/usage.md
```diff
@@ -925,11 +925,16 @@ proxy:
...
   # port: 11435          # Pin to override the mode default. Omitted = derived from
   # ollama_compat (11434 when true, 11435 when false).
-  # Loopback only — there is no `host` knob; LAN binding is
-  # a deferred follow-up.
+  # host: 0.0.0.0        # LAN bind (requires api_key unless insecure_no_auth).
+  # api_key: "..."       # Bearer token enforced whenever set.
+  # fallback_enabled: true   # Family-MRU fallback on auto-start failure.
+  # header_read_timeout_secs: 30
+  # idle_ttl_secs: 1800      # 0 disables idle eviction.
+  # max_body_size: 16777216  # Bytes; cap on every request body (default
```
deepu105: The wrap makes the second line read as its own YAML comment rather than a continuation of the first:

```yaml
  # max_body_size: 16777216  # Bytes; cap on every request body (default
  # 16 MiB; 0 rejects every non-empty body).
```

Either align the continuation under the inline comment, or shorten it so the key line stands on its own.

src/proxy/state.rs
```diff
@@ -128,6 +137,7 @@ impl ProxyState {
     ollama_compat: bool,
     fallback_enabled: bool,
     api_key: Option<String>,
+    max_body_size: usize,
```
deepu105, line 140: Nit, and not for this PR. `from_context_with_auth` is 5 positional params now, two of them adjacent bools, so `from_context(&ctx, false, true, cap)` would silently misbehave if `ollama_compat` and `fallback_enabled` ever got swapped. The bools predate this change. If the signature grows once more, an options struct is overdue.

deepu105 (commented):
Claude Opus 5: Went through this against the plan and ran the suite locally. Looks good, fine to merge from my side.

What I ran:

- `cargo test --features test-fixtures --test proxy_routing --test proxy_ui` : 21 passed
- `cargo test --features test-fixtures --lib -- proxy::route::tests config::loader::tests::proxy` : 23 passed
- `cargo clippy --all-targets --features test-fixtures -- -D warnings` : clean

The thing I was most worried about at plan review was the `from_context` call sites. The plan accounted for 6 of them, 5 test sites plus the production one in `daemon/mod.rs`. The real count is 18, and all 18 are updated here, including the 11 in `tests/` and the one in `benches/proxy_overhead.rs` that the plan never mentioned. I missed the bench one too when I reviewed the plan, so good catch.

Stale comment sweep is clean: `grep -rn "2 MiB" src/proxy/` now only hits the intentional `format_bytes` doc and test lines. Correcting myself on one thing here, the plan listed `forward.rs:86` as a stale 2 MiB comment but that was a false entry. It already read "the cap" rather than naming a number, so leaving `forward.rs` out of the diff was right.

Also good that the `usage.md` proxy YAML block got fixed on the way past. It was still claiming there was no `host` knob and no `api_key`, `tls_*` or fallback-tuning key, when all four have existed for a while. Pre-existing drift, glad it's gone.

One thing in the PR description I'd reword. It says two concurrent max-size requests "bound worst-case memory at 32 MiB", but nothing enforces two. There's no semaphore or connection limit anywhere in `src/proxy/`, so worst case is 16 MiB times however many bodies are in flight, and the default bump multiplies that by 8. On loopback that's a runaway-agent problem rather than an attack, but `proxy.host` LAN exposure is a supported mode, and there a buggy authenticated client can OOM the daemon a lot more easily than before. Nothing this PR needs to solve. I'd just phrase it as an illustration instead of a bound, and maybe add a line to `usage.md` saying the cap is per-request, not global.

The rest of my notes are inline. All nits, none of them block.

On the three points I raised at plan review that shipped as-is, I'm fine leaving them, but let's treat them as decisions rather than things to rediscover in six months:

- `0` means reject-all, not unlimited
- no upper bound on the value
- `max_body_size` vs the `_secs` / `_bytes` suffix every sibling key uses

The `0` one is now locked in by a test and documented in three places, so if it's ever going to move, pre-1.0 is the cheap moment.

src/proxy/route.rs
```diff
@@ -193,16 +198,19 @@ pub(crate) async fn buffer_body(body: Incoming, cap: usize) -> Result<Bytes, Bod
 /// Map a body-buffering [`BodyError`] to its proxy error response, so
 /// every surface that drains a request body (data plane + `/ui`) emits
 /// the same status + message for an oversized / unreadable / malformed
-/// payload.
-pub(crate) fn body_error_response(err: BodyError) -> ProxyResponse {
+/// payload. `cap` is the in-force limit — the 413 message names it as a
+/// human unit plus the exact byte count, so a cap sitting just under a
+/// unit boundary (e.g. 2 MiB − 1) can't read as the boundary itself.
```
deepu105: This isn't what the code does. `fmt_bytes` uses `{:.1}` for MiB, so 2 MiB - 1 still renders as the boundary. Probed against the real `init::detection::fmt_bytes`:

```
2097151 -> 2.0 MiB
2097152 -> 2.0 MiB
1048575 -> 1024 KiB
```

The human unit is exactly as rounded as the old `format_bytes` was. What makes the message honest is the `(2097151 bytes)` half, nothing else.

Worth rewording, because someone reading this comment has a reason to believe the byte count is redundant and drop it. Same claim in the plan's Decisions bullet ("no longer reads as the boundary itself") and in the E2E doc addendum.

src/proxy/route.rs
```diff
@@ -687,6 +695,50 @@ async fn collect_fallback_candidates(state: &Arc<ProxyState>) -> Vec<FallbackCan
...
+    // so a client sending exactly 2 MiB was told it "exceeds the 2 MiB
+    // limit" — a contradiction. The exact byte count keeps it honest.
+    let cap = 2 * 1024 * 1024 - 1;
+    let resp = super::body_error_response(super::BodyError::TooLarge, cap)
+      .expect("error response must build");
+    let collected = resp.into_body().collect().await.expect("body").to_bytes();
+    let v: serde_json::Value = serde_json::from_slice(&collected).expect("json envelope");
+    let message = v["error"]["message"].as_str().expect("message");
+    assert!(
+      message.contains(&format!("{} bytes", cap)),
```
deepu105: This can't fail for the reason it's named. The assertion only checks `2097151 bytes`, and that string comes straight from `cap` in the format call, never from the formatter. Swap `fmt_bytes` back for something that renders `2 MiB` and this test still passes, which is the exact regression it exists to catch.

`body_exceeding_cap_returns_413` in `tests/proxy_routing.rs` does it right by pinning both halves. Same thing here: assert the whole message, `request body exceeds the 2.0 MiB (2097151 bytes) limit`. That also pins the fact that the human half is deliberately rounded, which is the bit the comment above gets wrong.

docs/plans/2026-08-24-001-feat-proxy-max-body-size-plan.md
```diff
@@ -0,0 +1,71 @@
...
+## Design
+
+| Decision | Value |
+| --- | --- |
+| Key | `proxy.max_body_size` — **bytes**, `usize` (matches the issue's example; no human-size strings, no human-`status` surface) |
+| Default | **16 MiB** (`16 * 1024 * 1024`) — `BODY_LIMIT_BYTES` renamed `DEFAULT_BODY_LIMIT_BYTES`, becomes the serde default |
+| 0 | Legal, documented: rejects every non-empty body (a `Limited` cap of 0 trips on the first byte). No new error type, no validation pass |
+| Scope | One global cap for every body the proxy buffers — `/v1/*`, `/v1/messages*`, `/api/show`, `/ui` forwards |
+| Plumbing | `ProxyConfig.max_body_size` → `ProxyState.max_body_size` (set at daemon startup, like `fallback_enabled`) → threaded into `buffer_and_extract(body, cap)` and `body_error_response(err, cap)` |
+| 413 message | `request body exceeds the {cap} limit` with a small private `format_bytes` — largest unit the value fits in, whole when exact (`16 MiB`, `1 GiB`), two decimals otherwise with trailing zeros trimmed (`976.57 KiB`, never `2.00 MiB`); sub-KiB stays raw bytes (`512 B`) |
```
deepu105: This row still describes the private `format_bytes` that the follow-up commit deleted, two-decimal rules and the `976.57 KiB` example included. So the key-decisions table and the new Decisions section now say different things about the same message, and the Decisions section is the one that's right.

Plans are records and I'd rather not rewrite one after the fact, but this commit already edits the file, and the table currently documents code that isn't in the tree. A "superseded, see Decisions" note on the cell would do it.

deepu105 (commented):
Claude Opus 5: I reviewed `8d5a847`. The three things I flagged are properly closed:

- 413 now carries the exact byte count.
- `/ui` has a real cap test. Using a 1 KiB cap is the right call, since the default would let that body through, so it can only pass if the configured cap actually reaches `forward_ui`.
- The private `format_bytes` is gone and the canonical `fmt_bytes` is doing the work.

Locally: `proxy_routing` + `proxy_ui` 22 pass, the `route` / `loader` lib tests 23 pass, clippy clean on all targets.

Three small things left, all inline. Two of them are really the same mistake: `fmt_bytes` still rounds, so the "can't read as the boundary itself" claim isn't true of the human unit.

One thing that isn't in the diff: the PR description still says *"two concurrent max-size requests bound worst-case memory at 32 MiB"*. Nothing in the proxy enforces two, there's no concurrency or connection limit anywhere in `src/proxy/`, so worst case is 16 MiB x N in-flight bodies. `docs/usage.md` got the correct per-request wording in this commit but the description didn't, and the description is what ends up in the merge commit.

---

**On `0`**: I'd flip it to mean no check rather than reject-all. `idle_ttl_secs: 0` disables eviction in this same config block, nginx's `client_max_body_size 0` disables the check, and we already accept `99999999999` so effectively-unlimited is reachable regardless. Reject-all has no use case that `proxy.enabled: false` doesn't cover better, so `0` is currently spent on a mode nobody wants. It's one branch in `buffer_body` and pre-1.0 is the free moment. Detail in the `src/config/loader.rs` and plan-doc threads, including a correction to something I got wrong there (`--proxy-port 0` does not pick an ephemeral port).

src/config/loader.rs
deepu105: Coming back to the `0` question with a firmer answer: it should mean no check, not reject-all.

First, a correction on my own comment above. `--proxy-port 0` does not pick an ephemeral port. `effective_port` is just `port.unwrap_or(default)` with no zero handling anywhere in the CLI or config path, so that was a bad example. The other half holds up: `daemon/mod.rs:605` is `if opts.proxy.idle_ttl_secs > 0`, so `0` disables eviction in this same config block.

The precedent that matters most is nginx, same knob, same job. From its docs: *"Setting size to 0 disables checking of client request body size."*

And the safety argument for reject-all doesn't survive the other decisions we shipped. `max_body_size: 99999999999` already parses and is honored, deliberately. Effectively-unlimited is already reachable, just with an uglier spelling. Refusing the clean spelling for a capability we already grant isn't safer, it's just inconsistent.

Meanwhile reject-all has no use case I can find. `proxy.enabled: false` already kills the data plane and does it honestly (`status.proxy.status = "disabled"`) instead of 413ing every request. So `0` currently spends the one value users will actually reach for on a mode nobody wants.

Cost looks like one branch in `buffer_body`:

```rust
if cap == 0 {
  return body.collect().await.map(|c| c.to_bytes()).map_err(/* Read */);
}
```

plus flipping `proxy_config_max_body_size_zero_parses` into a behavioural test, and the three doc spots. Pre-1.0 is the free moment, after release it's a breaking change.

One caveat to document rather than imply: nginx's `0` is safe because nginx spools to disk. We buffer the whole body in RAM with no concurrency limit. So word it as "no cap, one request can buffer arbitrary RAM" rather than implying nginx-equivalent safety.

docs/plans/2026-08-24-001-feat-proxy-max-body-size-plan.md
```diff
@@ -0,0 +1,71 @@
...
+- `status.proxy.max_body_size` — config is not runtime state.
+- Human-size strings in config (`"16MiB"`) — raw bytes, per the issue.
+- Per-route caps (e.g. `/ui` separate from `/v1/*`).
+- Upstream (llama-server / ds4-server / lemond) body limits — verified against, never changed.
+
+## Decisions (pinned, not gaps)
+
+Raised at plan review, shipped as-is, and now locked in as decisions rather than open questions (PR #71 review):
+
+- **`0` means reject-all, not unlimited.** A `Limited` cap of `0` trips on the first byte, so `max_body_size: 0` refuses every non-empty body (HTTP 413). This is the opposite of how the sibling keys read (`idle_ttl_secs: 0` disables eviction, `--proxy-port 0` picks an ephemeral port), so a user reaching for `0` expecting "no limit" bricks their proxy. Deliberate: there is no "unlimited" mode, and `0` is the one value that can't be a real cap. Locked in by `proxy_config_max_body_size_zero_parses` and documented in `config.example.yaml` + `usage.md` + the `ProxyConfig` doc. If it ever moves, pre-1.0 is the cheap moment.
```
deepu105: This decision should flip: `0` should mean no check, not reject-all. Full argument in the `src/config/loader.rs` thread. Short version is that `idle_ttl_secs: 0` disables eviction in this same block, nginx's `client_max_body_size 0` disables the check, and we already accept `99999999999` so effectively-unlimited is reachable anyway. Reject-all has no use case that `proxy.enabled: false` doesn't cover better.

Separately, this bullet repeats a claim of mine that turned out to be wrong. `--proxy-port 0` does not pick an ephemeral port, `effective_port` is `port.unwrap_or(default)` with no zero handling. My error. It needs deleting from the bullet whichever way the decision goes.

deepu105:
Qwen 3.8 27B Q8:

## Fix summary — `max_body_size=0` flip + review follow-ups

Rebased onto current `main` and addressed the open review threads.

### `0` now means *no check*, not reject-all
- `buffer_body` gains a `cap == 0` branch that passes the body through `Limited` uncapped instead of tripping on the first byte.
- Matches the sibling keys (`idle_ttl_secs: 0` disables eviction) and nginx's `client_max_body_size 0` ("Setting size to 0 disables checking of client request body size").
- Caveat documented, not implied: nginx's `0` is safe because it spools to disk; we buffer the whole body in RAM with no concurrency limit, so `0` is worded as "no cap, one request can buffer arbitrary RAM," not nginx-equivalent.
- Locked in by `zero_cap_disables_the_body_check` (a 17 MiB body under cap 0 buffers fully → 503, not 413).

### 413 message honesty
- Reworded the doc comment + test asserts: the human half (`fmt_bytes`) still rounds at unit boundaries (2 MiB − 1 → `2.0 MiB`), so it is the **exact byte count** that keeps a sub-boundary cap honest, not the human unit. The whole message is now asserted, not just the byte-count substring.

### Docs
- `config.example.yaml`, `usage.md`, `architecture.md`, `ProxyConfig` doc, the plan's Decisions (superseded the reject-all bullet, dropped the wrong `--proxy-port 0` claim), and the E2E addendum.

### E2E (isolated daemon, umbrella pinned to 13399)
- cap 0 → 17 MiB body **not** 413 (reaches routing)
- cap 16 MiB (default) → same body **413**
- cap 2097151 (2 MiB − 1) → `2.0 MiB (2097151 bytes)`

Full suite + lint green.

deepu105:
All review comments aaddressed. Approved.

PR #73, opened 2026-08-31
fix: TUI Ctrl+P save-preset now carries backend and server

deepu105:
## Problem

`Ctrl+P` built `SavePresetCmd` from `model_path`/`name`/`knobs`/`extras` only, so the launch identity never reached `presets_save`. A preset saved from the TUI read back with `server: null` and silently fell to `servers[0]` instead of the build the model was actually running on.

For a fork-quantized GGUF that meant the preset pinned the **stock** binary rather than the **fork** build it was captured against. Found 2026-08-30 on a real `pi-coding` preset saved against the ROCmFP4 build.

The CLI path already sent identity explicitly (`src/cli/presets.rs`); only the TUI path was dropping it.

## Fix

- `SavePresetDialog` + `SavePresetCmd` gain `backend` / `server` (`Option<String>`).
- `open_save_preset_dialog` captures them from the running status row (honest resolved identity) when saving off a live launch, else from the picker's intended `model_backend.explicit_id()` + `selected_server`.
- `encode_writer_cmd` (now `pub`) emits both on the `presets_save` wire.
- Absent identity ⇒ `null`, so the daemon re-derives rather than storing a guess — mirrors the CLI "launch identity, not a knob" semantics.

## Tests

- New `the_tui_save_payload_carries_the_launch_identity` (`tests/knob_parity_test.rs`) pins the exact wire shape leaving the TUI — this is the gap the knob-parity suite had, which let the bug slip through.
- Extended `save_preset_from_running_launch_carries_dispatched_extras_and_identity` (`src/tui/app.rs`) for the running-launch capture path.
- Full lib 2229 passed; `tui::` 576 passed; `knob_parity_test` 6 passed.
- E2E: drove TUI `Ctrl+P` on a live isolated-daemon launch; the saved preset reads back `backend: llamacpp` / `server: llamacpp-rocm` (was `server: null`).

Closes `TODO.md` line 298.

---
🤖 **Authored by Qwen3.8-27B-Q8_0**


src/tui/app.rs
```diff
@@ -1842,17 +1842,28 @@ impl App {
...
+        m.server.clone(),
+      )
     } else if let Some(p) = &self.launch_picker {
       (
         p.user_knobs.clone(),
         p.extras
           .iter()
           .map(|s| s.to_string_lossy().into_owned())
           .collect(),
+        p.model_backend.explicit_id().map(str::to_string),
```
deepu105: `backend` is taken from `model_backend`, but that is only the model's routing/scoping backend. It can differ from the server selected in the picker: `resolved_backend()` switches to the selected server's `backend_id`, and normal launch sends `launch_backend()` with `selected_server`. For a ds4-compatible model with `model_backend=ds4` and `selected_server=llamacpp-rocm`, this saves `backend: ds4` with `server: llamacpp-rocm`; replay sees the explicit ds4 backend and can launch ds4 instead of the selected llama.cpp build. Please derive the saved backend from the selected server (or otherwise use the effective launch identity) so these fields cannot disagree.

src/tui/app.rs
```diff
@@ -1842,17 +1842,28 @@ impl App {
...
+    // rest, and the identity (which engine / build) pins the preset to the
+    // run it was captured from. A running row carries the *resolved*
+    // identity; a picker carries the *intended* one (its concrete engine and
+    // chosen server build).
+    let (knobs, extras, backend, server) = if let Some(m) = self.focused_managed() {
+      (
+        m.knobs.clone(),
+        m.extras.clone(),
+        m.backend.clone(),
+        m.server.clone(),
```
deepu105, line 1856: `m.server` comes from status `params.server`, but `--device` can select a non-default server by looking up the device owner in `launch_service` without writing that server id back into `LaunchParams.server`. Status then reports `null` even though the actual binary is, for example, a ROCm/Vulkan build, so Ctrl+P omits it and the saved preset later falls back to the default server. Please persist/stamp the device-derived server identity before status is built (or resolve it when capturing).

deepu105 (commented):
Found two medium-severity identity edge cases in the Ctrl+P capture path; details are inline.

src/tui/app.rs
```diff
@@ -1842,17 +1842,28 @@ impl App {
...
       (
         p.user_knobs.clone(),
         p.extras
           .iter()
           .map(|s| s.to_string_lossy().into_owned())
           .collect(),
+        p.model_backend.explicit_id().map(str::to_string),
+        p.selected_server.clone(),
       )
     } else if let Some(p) = self.build_default_picker() {
```
deepu105: The `launch_picker` branch (1858-1867) and this `build_default_picker()` branch (1868-1877) are identical bodies except for the source of `p` — both destructure the same 4-tuple. Worth extracting a shared helper so a future 5th field only has to be added once.

Separately, this branch does more work than it needs to: `build_default_picker()` also computes `native_ctx`, `mtp_capable_for`, `compatible_servers`, and re-resolves the full preset cascade via `effective_preset_choices` (lines 1592-1650) — none of which the destructured tuple uses. The two values it actually needs, `predicted_backend(&path)` and `last_params.get(&path)`, are already cheap and are exactly what `build_default_picker` itself pulls from internally (1603-1622, 1631-1633, 1641). Since `existing_preset_names` right after (line 1882) independently reruns the same catalog/preset-cascade resolution, a single Ctrl+P on the no-picker path currently resolves the preset cascade twice for no reason. Low severity (not a hot path), but a cheap fix alongside the dedup above.

src/tui/save_preset_dialog.rs
```diff
@@ -68,6 +75,8 @@ impl SavePresetDialog {
     model_name: String,
     knobs: crate::launch::knobs::KnobSet,
     extras: Vec<String>,
+    backend: Option<String>,
```
deepu105: `SavePresetDialog::open` was already under `#[allow(clippy::too_many_arguments)]` before this PR and now grows to 8 positional params, two of which — `backend` and `server` — are adjacent, same-typed `Option<String>`s. Nothing in the type system stops a transposed call if the single call site (`app.rs` `open_save_preset_dialog`) is ever reordered; a swap would silently pin the wrong backend/server with no compile error. The module already has a struct-of-args precedent for exactly this shape (`AppOptions` at app.rs:237, `StartModelArgs` at app.rs:452, the latter boxed for the same "don't bloat other variants" reason that would apply here). Worth migrating to a struct instead of continuing to grow the positional list.

src/tui/events.rs
```diff
@@ -2117,12 +2130,18 @@ fn encode_writer_cmd(cmd: WriterCmd) -> (&'static str, Value) {
...
+      // `server` pin the launch identity (which engine / build this preset
+      // was captured against) — the same identity the CLI's `presets save`
+      // sends, so a TUI-captured preset reproduces its run instead of
+      // silently falling to the daemon's default server.
       json!({
         "model_path": cmd.model_path,
         "name": cmd.name,
         "knobs": cmd.knobs,
         "extras": cmd.extras,
+        "backend": cmd.backend,
```
deepu105: This unconditionally serializes `backend`/`server` as explicit JSON `null` when unset (pinned by the new test asserting `bare_params.get("backend") == Some(&Value::Null)`). The CLI path sending the same `presets_save` RPC (`src/cli/presets.rs:178-183`) omits the key entirely when unset: `if let Some(b) = backend { payload.insert("backend", json!(b)); }`. Two clients now encode "no identity" for the same method two different ways. Works today only because serde_json's default `Option<T>` deserialization treats absent-key and explicit-null the same; if the daemon handler is ever changed to distinguish "leave existing field untouched" (absent) from "clear it" (null) — a common JSON-Patch-style need — these two paths silently diverge and neither side's tests would catch it, since each only checks its own encoding. Suggest matching the CLI's omit-when-None shape here for consistency.

deepu105 (commented):
Reviewed directly (diff read + targeted verification, no test failures found). One doc gap that isn't inline-commentable since the file isn't part of this diff: this is a user-visible bug fix (TUI-saved preset silently dropped `backend`/`server`, falling back to `servers[0]`) with no `CHANGELOG.md` entry under `[Unreleased]` — the project's own AGENTS.md "Docs ship with code" rule calls for one, and a near-identical sibling fix already has a bullet there.

deepu105:
Addressed all six review comments in `3204875`:

1. **Picker backend/server mismatch** — both fields now route through one `launch_identity()` helper (`src/tui/launch_picker.rs`). A non-default server pick determines its own backend (its owning backend), so the two can't disagree; with no non-default pick the backend falls back to the model's explicit choice (`None` for `Auto`).

2. **Device-derived server not stamped** — extracted `pick_launch_binary()` (`src/daemon/launch_service.rs`). When a `--device` selector resolves to a server, it now stamps that server's id (and, when the backend is still `Auto`, its owning backend) into `LaunchParams`, so status and Ctrl+P capture report the real build instead of `null`.

3. **Dedup + double cascade** — the open-picker and default-picker branches now share `picker_capture()`; the no-picker path drops its redundant `build_default_picker()` cascade and reads the identity straight from the cheap sources (`last_params` + `compatible_servers` + the predicted backend), since `existing_preset_names` reruns the cascade anyway.

4. **Struct of arguments** — `SavePresetDialog::open` now takes a `SavePresetArgs` struct, so the two adjacent `Option<String>` identity fields (`backend`, `server`) can't be transposed at the call site with no compile error.

5. **Omit-when-None** — the TUI's `encode_writer_cmd` now omits `backend`/`server` when unset (matching the CLI's `presets save`) instead of emitting explicit `null`.

6. **CHANGELOG** — added a `Fixed` entry.

**Tests:** added unit tests for `pick_launch_binary` (device stamping, explicit-server precedence, stale-selector drop) and updated the knob-parity test to assert the keys are *absent*, not `null`. Full suite green (lib + integration with `test-fixtures`), clippy `-D warnings` clean, fmt clean.

**E2E:** drove the daemon with an isolated state dir — saved a preset with `--backend llamacpp --server llamacpp-rocm` and read it back with both identity fields preserved; the no-identity save omits the keys on the wire.

src/tui/app.rs
```diff
@@ -1842,41 +1842,81 @@ impl App {
...
+      let knobs = last.map(|l| l.knobs.clone()).unwrap_or_default();
+      let extras = last.map(|l| l.extras.clone()).unwrap_or_default();
+      let selected_server = last.map(|l| l.server.clone()).unwrap_or_default();
+      let (backend, server) = crate::tui::launch_picker::launch_identity(
+        &selected_server,
+        &self.compatible_servers(&path),
+        &self
+          .predicted_backend(&path)
+          .map(crate::launch::params::BackendChoice::from_id)
+          .unwrap_or_default(),
```
deepu105: `.unwrap_or_default()` here falls back to `BackendChoice::Auto`, but `build_default_picker()` (line 1635), the function this branch is meant to mirror cheaply, falls back to `BackendChoice::from_id(DEFAULT_BACKEND_ID)` (explicit `llamacpp`) when `predicted_backend` is `None`. `predicted_backend` is just a `backend_by_path` lookup that stays empty until a `list_models` row carries a `backend` field, so for a model without one yet this path saves `backend: None` (omitted now, thanks to the omit-when-None fix above) where `build_default_picker` would pin `llamacpp` explicitly. On replay the daemon's own auto-routing decides the backend, which for a ds4-compatible GGUF can mean ds4 instead of the llamacpp build actually shown, the same class of bug this PR exists to fix, just reintroduced through this untested branch (`open_save_preset_dialog` has one test and it only covers the running-launch case, not this no-picker path). Please use the same `from_id(DEFAULT_BACKEND_ID)` fallback as `build_default_picker` here.

deepu105 (commented):
Found one more medium-severity identity edge case in the follow-up fix; details inline.

PR #74, opened 2026-09-01
feat: isolate preset launches from last_params and surface knob provenance

deepu105:
## What

Two related changes to how a launch resolves its knobs:

1. **Preset isolation** — a named preset (CLI `start --preset <name>` or a TUI form launch) is self-contained, but it still inherited a stale `last_params` through the `LastUsed` layer. `Explicit` now resolves as pure-fit and skips `LastUsed`, matching `--preset auto`. The inline-flag-only `start` path (no preset) is unaffected and keeps its `PresetDefault → LastUsed` fallback.

2. **Knob provenance** — the `start_model` IPC response and the CLI `--json` output now carry `layer_sources`, a map of each resolved knob to the layer its value came from (`user`, `preset_default`, `last_used`, `arch_default`, `model_default`, `server_default`). Omitted when empty (a pure-fit launch where every knob fell to the backend default).

## Why

A preset is meant to reproduce a whole run, not just how it was tuned. If a stale `last_params` leaked in, a preset launch could silently pick up knobs the preset never defined — e.g. `threads` from a previous inline launch. Provenance makes that visible and gives the TUI/CLI a way to show where each value came from.

## Verification

- `cargo test --features test-fixtures --lib` — 2230 passed, 0 failed
- New unit test `explicit_selection_is_pure_fit_and_skips_last_used` guards the gating
- E2E against a real `llama-server` + `Qwen3.5-4B` model:
  - Launch A (inline `--ctx-size 8192 --threads 4`) → `threads: user`
  - Launch B (`--preset mypreset`, which doesn't set `threads`) → `threads: server_default`, **not** `last_used` (before the fix it leaked `last_used`)
  - Launch C (no preset) → `ctx-size: last_used` (fallback preserved)

## Files

- `src/daemon/launch_service.rs` — `is_pure_fit` helper, `Explicit` gating, `layer_sources` threading, unit test
- `src/ipc/methods.rs` — `layer_sources` on the `start_model` response
- `src/cli/start.rs` — `layer_sources` on the `--json` output
- `src/cli/cli_args.rs` — `--json` help text
- `src/launch/knobs/def.rs`, `src/launch/params.rs` — serde derives for `KnobId` / `LayerLabel`
- `src/backend/lemonade/backend.rs` — thread `layer_sources` through
- `CHANGELOG.md`, `docs/architecture.md` — docs


deepu105 (commented):
Reviewed directly (diff read + traced the resolver + selection-gating logic against the pre-existing code, no test failures found). The preset-isolation half (`is_pure_fit` gating `LastUsed` at launch_service.rs:665-666) is correctly scoped — verified it's the only place `pure_fit` affects layer composition; extras/mtp/mmproj inheritance stay on the pre-existing `no_selection` gate, untouched by this change. One real issue on the provenance half below.

src/ipc/methods.rs
```diff
@@ -534,6 +534,11 @@ async fn start_model_handler(
   if !started.warnings.is_empty() {
     resp["warnings"] = json!(started.warnings);
   }
+  // Layer provenance per resolved knob. Omitted when empty (a pure-fit launch
+  // where every knob fell to the backend default resolves no real layer).
+  if !started.layer_sources.is_empty() {
```
deepu105, line 539: This `is_empty()` check can't actually be false for any real backend. `resolve_layered` (`src/launch/knobs/resolve.rs:90-91`, unmodified by this PR) seeds `sources` with a fallback label for *every* knob the backend declares, before any layer is applied:
```rust
for def in defs {
  sources.insert(def.knob_id(), def.fallback);
}
```
`seed_layerless` (resolve.rs:201-214) only touches `resolved.knobs`, never prunes `sources`. So `resolved.sources` — and therefore `started.layer_sources` (launch_service.rs:960, `layer_sources: resolved.sources`) — always has one entry per declared knob. llama.cpp alone declares dozens, so this map is never empty in practice.

Result: `layer_sources` is attached to *every* `start_model` response, including a full per-knob fallback dump on a pure-fit launch — the opposite of what every doc comment introduced in this diff claims (`StartedLaunch::layer_sources` here, and the matching comments in `cli_args.rs` and `cli/start.rs`: "omitted when empty ... a pure-fit launch ... resolves no real layer"). No test asserts the omitted-when-pure-fit case, which is presumably how this got through — `explicit_selection_is_pure_fit_and_skips_last_used` only tests `is_pure_fit` in isolation, not the response shape.

Either filter `sources` down to non-fallback entries before assigning `layer_sources` (e.g. drop entries where `label == def.fallback`), or the doc comments need to change to describe the actual always-present, full-knob-dump behavior — as written they document a code path that can't run.

deepu105:
Addressed the review comment and rebased onto `main`.

**`layer_sources` always present** — confirmed: `resolved.sources` is never empty because `resolve_layered` seeds a fallback label for every declared knob, so the field was attached to every `start_model` response (a full per-knob fallback dump even on a pure-fit launch) — the opposite of the documented omitted-when-empty behavior.

Fixed by filtering to the knobs a real layer actually supplied before assigning `layer_sources`, reusing the existing `is_layer_less` sentinel (`source == fallback`, exact since no real layer carries `ServerDefault`/`ModelDefault`). Added `Resolved::real_sources(backend_id)` for it; the full `sources` map is left intact for the TUI editor's origin chips.

Tests:
- unit: `real_sources_drops_knobs_that_fell_to_their_fallback` (empty on pure-fit; only real-layer knobs otherwise)
- integration: `auto_selection_start_inherits_nothing` now asserts the field is omitted on a pure-fit launch

Rebased onto `main` (`00e3ed0`); resolved the `CHANGELOG.md` and `launch_service.rs` conflicts by keeping both sides' additions. Full suite + clippy green.

deepu105:
Added the missing end-to-end regression coverage for the namesake scenario.

**Gap:** the only guard was a unit test on `is_pure_fit()` in isolation; the leak-prevention itself was verified only by the manual E2E in the PR description, not by CI.

**New test** — `named_preset_launch_does_not_inherit_stale_last_params` (`tests/start_model_ipc_test.rs`):
1. Inline launch persists `threads=4` into `last_params`.
2. A `selection: explicit` launch (preset sets `ctx-size`, not `threads`) must not inherit it from the `LastUsed` layer.
3. Asserts the preset's own knob resolves from the `user` layer (proves the preset actually applied — guards against a false pass where nothing resolved) and that `threads` is **not** sourced from `last_used`.

Verified it fails when the `Explicit` pure-fit gate is reverted (`threads` comes back as `last_used`), and passes with the fix. Full suite green (2235 lib + 12 integration), clippy clean.

PR #75, opened 2026-09-03
feat: run alias and YAML launch files for start

deepu105:
## Summary

`run` is now a visible alias for `start`. Same command, same flags, nothing mode-like about it. It exists because "run this model" is what people actually want to type.

The positional also takes a `.yaml`/`.yml` launch file now:

```bash
llamastash run qwen3.8.yml                 # the file's default preset
llamastash run qwen3.8.yml --preset slow   # a named entry from the file
llamastash run qwen3.8.yml --ctx 4096      # flags still layer on top
```

The file is a `presets:` block in `config.yaml`'s own shape, narrowed to one model. No new preset types, and it writes nothing back. The point is you can hand someone a file that says "this model, these settings, launch it" without touching their `config.yaml`.

```yaml
presets:
  Qwen3-8B-Q4_K_M.gguf:      # substring / exact name / catalog path
    default: fast
    entries:
      fast:
        knobs:
          n_gpu_layers: 99
          ctx_size: 32768
        backend: llamacpp
      slow:
        knobs:
          n_gpu_layers: 40
```

It is not self-contained, on purpose. The model key resolves against the running daemon's catalog exactly like a `start` argument, so a key that matches nothing exits `66`. Detection reads the positional and not the alias, so `start file.yml` behaves identically, and a model actually named `foo.yml` that isn't on disk still resolves as a model reference.

**Validation is stricter than `config.yaml`, and that is the part worth reading.** A launch file is hand-authored by definition and has no inspection surface, you never see its parsed form. Today `config.yaml` drops an undeclared knob with a `log::warn!` that only reaches the terminal under `--verbose`, which is fine there: the daemon parses it once and `presets show` can display the result. In a launch file, `n_gpu_layerz: 99` would launch a configuration that reads differently than it runs, with nothing on screen. So each of these exits `64` with the problem named:

| The file | Result |
| --- | --- |
| no `presets:` entries, or 2+ model keys | must name exactly one model |
| a model key with no `entries:` | must define at least one preset |
| 2+ entries, no `default:`, no `--preset` | ambiguous, the message names the entries |
| `default:` naming no entry | dangling default (`config.yaml` ignores this one silently) |
| `--preset auto`, or `default: auto` | `auto` means "apply no preset", which a launch file cannot mean |
| a knob no backend declares | names it and points at `llamastash knobs` |

Only the selected entry's knobs get checked. An undeclared knob in an entry you didn't launch stays silent.

**Reuse over a new launch path.** The file's preset goes through `materialize_preset` -> `LaunchParams::to_wire` -> `partial_params_from_preset`, the same projection `--preset <name>` already uses, and reports `selection = "explicit"` so the daemon skips both the default-preset and last-used layers. A hand-written field-for-field converter was the alternative, and that's the drift bug you get for free: the comments in that code already record two fields dropped this way (`backend:`, a disarmed `mtp:`). `launch_params_row` moved out of `ipc::methods` to become `LaunchParams::to_wire` for this, since it's the wire projection of a launch type and belongs beside it, same as `CatalogRow::to_wire_value`.

Everything from `resolve_mode` down is untouched: `--ctx`, `--port`, the knob overlay, `build_payload`, `start_model`. Extras keep their existing asymmetry, flags after `--` replace the file's whole `extras:` list instead of merging into it, exactly as they do for `--preset` today.

## Test plan

- [x] `cargo test --features test-fixtures` (2277 lib tests plus every integration binary)
- [x] `make lint` (`fmt --check` + `clippy --all-targets -- -D warnings`)
- [x] 7 new E2E tests spawn the real binary against a per-test daemon on the `fake_llama_server` fixture, so the clap alias and `--json` stdout are genuinely exercised instead of asserted around: alias launch, `start`/`run` body parity, the file's `default:` preset, `--preset` selecting from the file, flag layering plus the extras replacement, `66` on an unresolvable key, `64` on two models
- [x] All 8 validation rows driven by hand against the real binary on an isolated daemon (temp state/config/cache, port 21435, `--no-spawn`), each printing the message in the table above
- [x] `run --help` is byte-identical to `start --help`

Docs ship in the same commit: a Launch files section in `docs/usage.md`, a README quickstart line, one CHANGELOG entry.

Implemented by Qwen3.8-Flash-Next-UD-Q4_K_XL.


src/cli/launch_file.rs
```diff
@@ -0,0 +1,391 @@
...
+    .and_then(|v| v.get(&sel.preset_name))
+    .and_then(|v| v.get("knobs"))
+    .and_then(|v| v.as_mapping())
+  else {
+    return Ok(());
+  };
+  let mut unknown: Vec<String> = knobs
+    .iter()
+    .filter_map(|(k, _)| k.as_str())
+    .filter(|k| crate::launch::knobs::resolve_id(k).is_none())
```
deepu105: This checks the knob **id** and stops there, but the deserializer drops a bad **value** on a good id in exactly the same silent way. `coerce()` in `src/launch/knobs/serde_impl.rs` logs ``knob `{id}`: {e}; ignoring`` and returns `None`, and that log line is just as invisible to the CLI as the unknown-knob one you're closing here.

So the file still does the thing the description says it must not:

```yaml
entries:
  fast:
    knobs:
      n_gpu_layers: 99
      ctx_size: 8k
```

Against the branch binary:

```
exit=0  preset=fast
last_params: {"ctx":null,"knobs":{"n-gpu-layers":99,"reasoning":false},...}
```

The 8k context is gone. Nothing on screen, exit 0. That's the `n_gpu_layerz` case wearing a different typo, and I'd argue it's the more likely one to type.

Cheap fix that catches both in one pass: you already have the parsed set in `sel.body.knobs`, so instead of only asking `resolve_id(k).is_some()`, check that each raw key both resolves *and* survived into `sel.body.knobs`. A key that resolved but isn't in the parsed set is precisely a rejected value.

src/cli/launch_file.rs
```diff
@@ -0,0 +1,391 @@
...
+fn reject_unknown_knobs(raw: &yaml_serde::Value, sel: &LaunchFileSelection) -> Result<(), CliExit> {
+  let Some(knobs) = raw
+    .get("presets")
+    .and_then(|v| v.get(&sel.model_key))
+    .and_then(|v| v.get("entries"))
+    .and_then(|v| v.get(&sel.preset_name))
+    .and_then(|v| v.get("knobs"))
+    .and_then(|v| v.as_mapping())
+  else {
+    return Ok(());
```
deepu105, line 195: This bails to `Ok(())` whenever the raw path doesn't resolve, so any shape it can't navigate skips validation altogether instead of failing. YAML merge keys land right in it, and merges are the first thing you reach for when a file has two entries sharing most of their knobs:

```yaml
presets:
  m.gguf:
    default: slow
    entries:
      fast: &base
        knobs:
          n_gpu_layers: 99
          n_gpu_layerz: 1
      slow:
        <<: *base
```

`run merge.yml` on the branch binary:

```
exit=0  preset=slow
last_params: {"knobs":{"reasoning":false},"extras":[],"ctx":null}
layer_sources: {"reasoning":"user"}
```

Two failures stacked. `yaml_serde` doesn't apply merge keys unless you call `apply_merge()`, so `slow` parsed as an empty `PresetBody` and launched a completely empty configuration while happily reporting `preset: "slow"`. And since `entries.slow.knobs` doesn't exist in the raw doc, the `n_gpu_layerz` riding in through the anchor was never looked at either.

Either call `apply_merge()` on `raw` before both parses and let merges work, or reject a `<<` key outright with the usual 64. Silently launching nothing is the one option that shouldn't survive, especially in the file type whose whole pitch is that it can't misreport itself.

Separately, I'd make this `else` branch fail rather than pass on principle. A validator that can't find what it's validating hasn't validated anything.

src/cli/launch_file.rs
```diff
@@ -0,0 +1,391 @@
...
+use serde::Deserialize;
+
+use crate::cli::exit_codes::{CliExit, USAGE};
+use crate::config::{ConfigPresetBlock, PresetBody};
+use crate::launch::presets::AUTO_DEFAULT;
+
+#[derive(Debug, Deserialize)]
+struct LaunchFileDoc {
+  #[serde(default)]
+  presets: BTreeMap<String, ConfigPresetBlock>,
```
deepu105, line 21: Same argument as the knob check, one level up. Neither `PresetBody` nor `ConfigPresetBlock` sets `deny_unknown_fields`, so a misspelled non-knob field just evaporates:

```yaml
entries:
  fast:
    knobs:
      n_gpu_layers: 99
    backendd: llamacpp
    extra: ["--rope-freq-base", "1000000"]
    mode: embedding
```

exit 0, and all three are gone: ran on the default backend, `extras: []`, mode chat.

`backend:` and `server:` are launch identity, not tuning. A typo there is worse than a typo'd knob, not better, because it silently picks a different engine and still reports success.

The `mode:` line isn't a strawman. `docs/architecture.md:393` still says "Alongside its `knobs:` map an entry may pin `mode:`, `backend:` and `server:`", which was true before mode became a knob. Anyone writing a launch file from that sentence gets it dropped without a word. That doc line is wrong on `main` too, so worth a separate fix either way.

You probably can't just add `deny_unknown_fields` to `PresetBody`, since that would change `config.yaml` loading for everyone. A launch-file-local mirror struct, or a raw-doc key check next to the knob one, keeps it scoped to the file type that actually needs it.

docs/usage.md
```diff
@@ -358,6 +358,53 @@ Every knob has three states:
...
+llamastash run qwen3.8.yml                 # the file's `default:` preset
+llamastash run qwen3.8.yml --preset slow   # a named entry from the file
+llamastash run qwen3.8.yml --ctx 4096      # flags still layer on top
+```
+
+The file is a `presets:` block in `config.yaml`'s own shape, narrowed to one model — no new preset types:
+
+```yaml
+presets:
+  Qwen3-8B-Q4_K_M.gguf:      # model key: substring / exact name / catalog path
```
deepu105, line 375: "substring / exact name / catalog path" is accurate for what the code does, but this is the one spot where the file isn't "`config.yaml`'s own shape".

In `config.yaml` a top-level key that doesn't name a discovered model is read as an **arch id** (`docs/architecture.md` § Named presets). Here the key goes straight to `resolve_model_with_candidates`, which is catalog-only. So an arch-keyed block copied out of someone's config either exits 66 with "no model matches", or substring-matches a pile of models and exits 66 with an ambiguity list. Both are confusing when the same block works fine in `config.yaml`.

I'm not asking for arch support in launch files, that's a fair scope cut. I'd just say it here: the key resolves like a `start` argument, and arch keys don't carry over. Copying a block out of `config.yaml` is the obvious first thing anyone will try.

src/cli/start.rs
```diff
@@ -33,27 +33,56 @@ use crate::launch::knobs::{Concept, KnobSet};
 pub async fn handle(args: StartArgs, cli: &Cli, config: &Config) -> CliResult {
   let mut client = connect_or_spawn(cli, config).await?;
```
deepu105: Nit, but an easy win: `connect_or_spawn` and `fetch_catalog` both run before the launch file is even read, so `llamastash run typo.yml` spawns a daemon and pulls the catalog only to exit 64 on a file it could have rejected on its own.

`load()` needs nothing from the daemon. Only `select_start_row` does. Hoisting the `is_launch_file` / `load` block above line 34 makes every pure-usage error in the validation matrix free, which also makes them nicer to test.

deepu105 (commented):
Opus 5: Went through this without trusting the description, built the branch, and ran the validation matrix against the real binary on an isolated daemon (temp state/config/cache, `--no-spawn`, `fake_llama_server`) instead of reading it.

The happy path holds up. I checked that a file's `mode` knob, its `extras` and its knob values genuinely reach the launch, and the `materialize_preset` -> `to_wire` -> `partial_params_from_preset` reuse does what you say it does. Moving `launch_params_row` onto `LaunchParams::to_wire` is the right call and the field-parity unit test is the right test to have written.

My problem is the validation section, which you flag as the part worth reading. The goal there is that a hand-authored file can never launch a configuration that reads differently than it runs, since there's nothing to inspect afterwards. Right now it closes one of the three ways that happens, and it fails open on a fourth. All four below exit `0` with nothing on stderr:

| launch file says | what actually launched |
|---|---|
| `ctx_size: 8k` | `ctx: null`, knob gone. Bad *value* on a good id drops as silently as a bad id |
| `backendd: llamacpp` | dropped, ran on the default backend |
| `mode: embedding` (entry level) | dropped, ran as chat |
| `slow: {<<: *base}` | empty knob set, `preset: "slow"` reported, and the bad knob id inside the anchor never checked |

The knob id check is good, it just guards one of four doors. Details inline.

One doc thing on top of that: `docs/architecture.md` § Named presets is the authority on presets per AGENTS.md, and it doesn't know launch files exist. Given the project rule that docs ship in the same commit, I think that section needs a line saying a launch file is a read-only preset source that never writes.

src/cli/launch_file.rs
deepu105: > check that each raw key both resolves *and* survived into `sel.body.knobs`. A key that resolved but is not in the parsed set is precisely a rejected value.

Done, exactly that. `validate_selection` splits the raw knob keys into three buckets: unresolved id, resolved-but-absent (rejected value), and fine. `ctx_size: 8k` now exits 64 with `sets knob values the backend rejects: ctx_size` and points at `llamastash knobs` for the type. Unit test `a_bad_value_on_a_good_knob_id_is_usage`, E2E `run_launch_file_with_a_bad_knob_value_exits_64`.

One case worth naming since it lands in the same bucket: `n_gpu_layers:` with an empty value, and `some_knob: auto` where that knob has no Auto state, both drop through `coerce` too, so both now error instead of launching without them.

src/cli/launch_file.rs
deepu105: > Either call `apply_merge()` on `raw` before both parses and let merges work, or reject a `<<` key outright with the usual 64.

Went with `apply_merge()`, and took it one step further: there is now a *single* parse. `raw` is parsed from text, merged, and the typed doc comes from `yaml_serde::from_value(raw.clone())` instead of a second `from_str`. That makes it structurally impossible for the validator and the deserializer to disagree about what the file says, which was the underlying problem here rather than merges specifically. Probed the crate first: `from_str` leaves `<<` in the tree (0 knobs for `slow: {<<: *base}`), `apply_merge` expands it, and `from_value` on the merged tree gives 1 knob plus the pinned backend.

> I would make this `else` branch fail rather than pass on principle.

Done, both of them. A `presets:` entry or an `entries:` entry that is not a mapping is now a 64 naming the model or preset. Since the typed doc parsed as `BTreeMap<String, ConfigPresetBlock>` by that point, a non-mapping there cannot occur on a file that would otherwise launch, so failing closed costs no valid file.

Your stacked case is covered on both halves: `a_bad_knob_inside_a_merged_anchor_is_usage` catches the `n_gpu_layerz` riding in through the anchor, and `run_launch_file_merge_keys_launch_the_merged_config` asserts the merged knobs actually reach `last_params`. Documented that merges are shallow per key, so an entry with its own `knobs:` replaces the anchor rather than extending it.

src/cli/launch_file.rs
deepu105: > A launch-file-local mirror struct, or a raw-doc key check next to the knob one, keeps it scoped to the file type that actually needs it.

Took the raw-doc key check, since the merged tree is already in hand. `BLOCK_FIELDS` (`default`, `entries`) and `ENTRY_FIELDS` (`knobs`, `extras`, `backend`, `server`) are checked against the selected entry, shared `PresetBody` / `ConfigPresetBlock` untouched, so `config.yaml` loading is unchanged.

> `backend:` and `server:` are launch identity, not tuning.

Agreed, and it drove the message design: `preset \`fast\` has fields no preset block reads: backendd, extra` followed by `a preset takes: knobs, extras, backend, server`.

> The `mode:` line is not a strawman.

Right, and the hint handles it generically rather than special-casing the name: if an unknown key is itself a knob id, the message adds `knobs go under \`knobs:\`, not beside it`. That catches `mode:`, a bare `ctx:`, `gpu_layers:` written at entry level, all of them.

> That doc line is wrong on `main` too, so worth a separate fix either way.

Fixed here rather than leaving the contradiction: `docs/architecture.md` now says an entry may pin `backend:` and `server:`, and that `mode` is a knob riding inside `knobs:`. The `mode` rung in the daemons

deepu105: > A launch-file-local mirror struct, or a raw-doc key check next to the knob one, keeps it scoped to the file type that actually needs it.

Took the raw-doc key check, since the merged tree is already in hand. `BLOCK_FIELDS` (`default`, `entries`) and `ENTRY_FIELDS` (`knobs`, `extras`, `backend`, `server`) are checked against the selected entry. Shared `PresetBody` / `ConfigPresetBlock` untouched, so `config.yaml` loading is unchanged.

> `backend:` and `server:` are launch identity, not tuning.

Agreed, and it drove the message: `preset `fast` has fields no preset block reads: backendd, extra` followed by `a preset takes: knobs, extras, backend, server`.

> The `mode:` line is not a strawman.

Right, and the hint handles it generically instead of special-casing the name: when an unknown key is itself a knob id, the message adds `knobs go under `knobs:`, not beside it`. That catches `mode:`, a bare `ctx:`, and `gpu_layers:` written at entry level, all of them.

> That doc line is wrong on `main` too, so worth a separate fix either way.

Fixed here rather than leaving the contradiction in place. `docs/architecture.md` now says an entry may pin `backend:` and `server:`, and that `mode` is a knob riding inside `knobs:`. The `mode` rung in the daemon's no-selection resolution is untouched, since that part still holds.

docs/usage.md
deepu105: > I'd just say it here: the key resolves like a `start` argument, and arch keys don't carry over.

Added, and it names the failure so a reader recognises their own copy-paste: arch keys do not carry over, a block copied out of a config under `qwen3` exits `66` (no match, or an ambiguity list) rather than applying to every Qwen, name one model. Scope cut kept as you framed it, no arch resolution in launch files.

src/cli/start.rs
deepu105: > Hoisting the `is_launch_file` / `load` block above line 34 makes every pure-usage error in the validation matrix free, which also makes them nicer to test.

Hoisted. `load` now runs before `connect_or_spawn` and `fetch_catalog`, so the whole validation matrix is daemon-free. Confirmed against the binary: four bad files under `--no-spawn` against an empty state dir each printed their message and left no `runtime.json` behind, so nothing spawned.

That bought a test that was awkward before: `run_launch_file_rejects_a_bad_file_with_no_daemon_running` drives a typo'd file with no daemon at all and asserts `64` on the knob name rather than a connection error.

deepu105:
Qwen3.8-Flash-Next-UD-Q4_K_X: Addressed in 9e8e5b1. All four doors closed, plus the doc gap.

The framing I took: this was not four bugs, it was one missing invariant. The validator was asking "is this id declared" when the question is "does this key reach the launch". So there is now a single parse, merge keys applied to it, and the typed doc deserialized from that same merged tree rather than a second `from_str`. The validator and the deserializer cannot disagree about what the file says, which is the property the whole strictness argument depends on. Each raw key in the selected entry is then checked against what survived: unknown block/entry field, undeclared knob id, or resolved id whose value was rejected.

| launch file says | now |
|---|---|
| `ctx_size: 8k` | `64`, `sets knob values the backend rejects: ctx_size` |
| `backendd: llamacpp` | `64`, `has fields no preset block reads: backendd` + what a preset takes |
| `mode: embedding` beside `knobs:` | `64`, same, plus `knobs go under `knobs:`, not beside it` |
| `slow: {<<: *base}` | merges work, anchor's knobs reach `last_params`, and a typo inside the anchor is caught |

Two things your comment made me do rather than the literal fix:

- **Both** un-navigable raw paths fail now, not just the merge one. Failing closed costs no valid file, because by that point the typed doc has already parsed as `BTreeMap<String, ConfigPresetBlock>`, so a non-mapping there cannot occur on a file that would have launched anyway.
- The `mode:` hint is generic rather than a special case for that name. If an unknown key is itself a knob id, the message points at `knobs:`. That catches a bare `ctx:` or `gpu_layers:` at entry level too, which is the shape someone migrating an old config will write.

Also fixed the stale `docs/architecture.md:393` sentence you spotted, since leaving it would keep generating the `mode:` case: an entry pins `backend:` and `server:`, `mode` is a knob inside `knobs:`. The daemon's `mode` rung in no-selection resolution is untouched, that part still holds. § Named presets now records the launch file as a read-only preset source, its two differences from the config (catalog-only keys, stricter validation), and that it shares the `materialize_preset` projection.

Coverage: 7 unit tests and 4 E2E on the new doors, including the stacked merge-plus-typo case and one that runs a bad file with no daemon at all. `cargo test --features test-fixtures` 2590/2590, clippy and fmt clean.

One correction to my own PR body: it claimed `make lint` covers `fmt --check`. It is clippy only, fmt runs as a separate CI job. Both are clean here, and the pre-commit hook does run `cargo fmt --all`.

deepu105:
Opus 5: Approved

All five findings verified fixed against the fix commit 9e8e5b1. I re-ran every original repro on the real binary in your .worktrees/feat-run-alias-yaml-launch checkout

Verdict: ship it. Two non-blocking follow-ups, neither worth holding the PR:

1. BLOCK_FIELDS / ENTRY_FIELDS (launch_file.rs:157-158) are hardcoded string arrays with no compile-time or test link to ConfigPresetBlock / PresetBody. Add a field to PresetBody and launch files reject it with "has a field no preset block reads", and nothing catches that. Same drift class the PR description calls out avoiding with the to_wire reuse. The failure is a loud 64 rather than a silent drop, so it's a follow-up, not a gate.
2. The rejected-value message names the knob but not the reason (ctx_size, not ctx_size: 8k is not an integer). coerce() has the actual parse error and throws it away into the log. Minor.

PR #76, opened 2026-09-04
feat: named launches — model@name addressing across proxy, CLI, TUI

deepu105:
## Summary

Allows multiple concurrent launches of the same model, each addressable by a user-chosen name via `model@name` syntax.

**Proxy:** `POST /v1/chat/completions` with `"model": "qwen3@coder"` routes to the named launch. `GET /v1/models` publishes named ids alongside catalog ids.

**CLI:** `llamastash start <model> --name coder`, `llamastash stop coder`, `llamastash show coder`, `llamastash list` shows a NAME column.

**TUI:** `Alt+Enter` on a model row opens a launch-name dialog. Named launches show their name in the list pane.

## Key changes

- `route.rs` / `router.rs`: parse `@name` suffix, filter supervisor walk by name, publish named ids in `/v1/models`
- `launch_service.rs`: stamp `StartParams.name`, refuse duplicate named launches, `drop_running_snapshots` on stop
- `coalesce.rs`: flight key extended to `(ModelId, Option<String>)` so concurrent cold requests for different names don't coalesce
- `launch.rs`: `auto_start` sets `name` but not `force`
- `launch_name_dialog.rs` (new): TUI dialog for naming a launch
- `keybindings.rs` / `events.rs`: `Alt+Enter` binding
- CLI: `--name` flag on `start`, name-aware `stop`/`show`/`list`/`resolve`

## Tests

- D2 fail-safe: model file with `@` in name resolves as plain reference
- D4: named request auto-starts carrying the name
- Two named launches of same model land on different ports
- `/v1/models` lists named row while live, drops after stop
- Concurrent cold requests for two names produce two launches
- TUI: `Alt+Enter` opens dialog, plain `Enter` does not
- All 2345 existing tests pass; clippy + fmt clean

## Plan

`docs/plans/2026-09-03-002-feat-named-launches-plan.md`


src/proxy/route.rs
```diff
@@ -249,21 +255,39 @@ pub(crate) async fn decide(state: &Arc<ProxyState>, body_model: Option<String>)
...
+  // contains `@` (e.g. `foo@bar.gguf`) resolves as a plain model reference.
+  // Only when the whole string does not resolve do we split on `@` and treat
+  // the right side as a launch name. When present, the launch must match
+  // both the model (path) and the name; the name is threaded through to
+  // `auto_start` so a second named launch of the same model gets its own
+  // flight and its own addressable id.
+  let (name, resolved) = match resolve_model_with_candidates(&rows, &requested) {
+    Ok(r) => (None, r),
+    Err(_) => {
+      let (m, n) = match requested.split_once('@') {
```
deepu105: This is the second implementation of the same `model@name` parse. `resolve_running` in `src/cli/resolve.rs` also splits on the first `@`, but with no empty-name guard and no whole-string-first fail-safe, and it compares the name lowercased while this path is case-sensitive. So the copies have already drifted on three axes.

Both also split on the first `@` while D2 says split on the last, so `foo@bar.gguf@coder` misparses in both places.

I'd add one `parse_named_reference()` next to `resolve_model_with_candidates` in `src/launch/resolve.rs`, which the CLI and proxy already share, and call it from both sites. One rule, and the D2 fix becomes a one-line change.

src/proxy/router.rs
```diff
@@ -363,6 +365,20 @@ async fn list_models(state: Arc<ProxyState>) -> ProxyResponse {
...
     .map(|m| ModelObject::new(published_id(&ids, m)))
     .collect();
+  // Named launches of a model that's already in the catalog get their own
+  // addressable id of the form `<model-id>@<name>`. A second launch of the
+  // same model with a name is reachable only through this id.
+  let state_snap = state.ctx.state.snapshot().await;
+  for r in &state_snap.running {
+    if let Some(name) = r.name.as_deref() {
+      let base_id =
+        crate::util::paths::model_public_id(std::path::Path::new(&r.params.model_path), None);
```
deepu105: This block is copy-pasted in `ollama_tags` further down: same snapshot, same `model_public_id(path, None)`, same `format!("{base_id}@{name}")`, same linear dedup scan. Only the `TagModel` dressing differs.

The bigger issue is that it bypasses `published_id(&ids, ...)`, used ten lines above. Two same-named `qwen3.gguf` in different roots publish bare `qwen3@coder` instead of the disambiguated stem, and a request for that id then 400s as ambiguous on the split. D6 said published_id_index stays the single rule for the model half.

Extract one `named_ids()` helper and run it through the same published-id machinery both listings already use.

src/proxy/route.rs
```diff
@@ -281,10 +305,27 @@ pub(crate) async fn decide(state: &Arc<ProxyState>, body_model: Option<String>)
...
   for (_launch_id, model) in sup_snap.into_iter() {
     if !same_path(&model.id().path, &resolved.path) {
       continue;
     }
+    // When a name is present, only a launch with that name is a match.
+    if let (Some(n), Some(st)) = (&name, &state_snap) {
+      let has_name = st
+        .running
+        .iter()
+        .any(|r| r.port == model.port() && r.name.as_deref() == Some(n.as_str()));
```
deepu105: This port-to-name join is a copy of the one in `attach_target` (`src/proxy/launch.rs`), comments included. Worth extracting one `launch_carries_name(snap, launch, name)` predicate.

While at it, the join should key on `launch_id` rather than the port. `ipc/status.rs` documents that a port is reused the moment its launch stops, and both loops already destructure the launch id into `_launch_id`. Extracting the helper means fixing the join once instead of twice.

src/daemon/launch_service.rs
```diff
@@ -310,6 +320,22 @@ pub(crate) async fn compose_and_spawn(
...
   }
+  // A name is unique per model: a second launch of the same model with the
+  // same name is refused so `<model-id>@<name>` stays a stable address.
+  if let Some(name) = parsed.name.as_deref() {
+    let model_path = parsed.model_path.clone();
+    let state_snap = ctx.state.snapshot().await;
+    let dup = state_snap
+      .running
+      .iter()
+      .any(|r| r.params.model_path == model_path && r.name.as_deref() == Some(name));
```
deepu105: Third variant of the same question, "does this running row carry this name": here it joins on `model_path ==` and is case-sensitive, both proxy walks join on `port ==` and are case-sensitive, and the CLI resolver compares lowercased. Net effect: `stop Qwen3@CODER` resolves fine while a proxy request for the same string falls through to auto-start and the gate passes, so one case variant launches a second full copy.

One canonical predicate shared by the gate, both proxy walks, and the CLI resolver would pin a single rule. Separate behavior question on the same line: the gate matches rows in any state, so a launch stuck in `error` keeps its name locked until it is stopped manually.

src/proxy/coalesce.rs
```diff
@@ -61,11 +61,22 @@ enum SlotState {
   Cancelled,
 }
 
+/// The single-flight key: the canonical model id plus the optional
+/// launch name, so two named launches of the same model don't coalesce
+/// into one flight.
+type FlightKey = (ModelId, Option<String>);
```
deepu105, line 67: Nit: `FlightKey` is declared here but `Leader::key` and `acquire` spell out the tuple. Use the alias so the key shape has one name.

src/daemon/state_store.rs
```diff
@@ -146,6 +146,11 @@ pub struct RunningSnapshot {
   /// `None` only on a legacy/adopted row that predates the stamp.
   #[serde(default, skip_serializing_if = "Option::is_none")]
   pub launch_id: Option<LaunchId>,
+  /// User-chosen name for this launch. When set, the launch is
+  /// addressable by `<model-id>@<name>` in `stop`, `logs`, and the
+  /// proxy's `body.model`. `None` for unnamed launches (the default).
+  #[serde(default, skip_serializing_if = "Option::is_none")]
+  pub name: Option<String>,
```
deepu105, line 152: Nit: adding this field meant hand-editing roughly 15 `name: None` literals across 8 test modules. The `launch_service` tests got a `named_running()` builder, the rest got nothing. A small snapshot test builder would absorb the next field addition too.

src/tui/launch_name_dialog.rs
```diff
@@ -0,0 +1,158 @@
+//! `Alt+⏎` "launch as…" modal: name the focused model before launching it.
```
deepu105: Nit: this is a structural clone of `save_preset_dialog` minus one stage, and the hint label already drifted: it resolves `LaunchNamed` under `Focus::ConfirmPopup`, but the action is scoped `FocusSet::LIST`, so the lookup always misses and user keybinding overrides never show in this dialog. Fine to leave as is for now, but if a third single-field modal appears, extract the shared frame (centered rect, panel, input, hint line) instead of cloning again.

deepu105 (commented):
DRY pass on the new named-launch code. Four real duplications, and each one has either already drifted between its copies or points straight at a routing bug, so deduplicating first is the cheapest way to fix the correctness cluster:

- `model@name` parsing exists twice, proxy and CLI resolver, and the copies disagree on empty names, case, and the whole-string fail-safe
- named-row id emission is copy-pasted between `/v1/models` and `/api/tags`, and bypasses `published_id` in the process
- the port-to-name join is copy-pasted between `decide()` and `attach_target()`
- "does this running row carry this name" is written three different ways with three comparison semantics

Three smaller nits inline as well.

src/tui/keybindings.rs
```diff
@@ -798,6 +802,11 @@ fn build_default_bindings() -> Vec<Binding> {
     hint: "launch", description: Some("launch focused model"),
     chords: [(KeyCode::Enter, KeyModifiers::NONE, ENTER_LABEL, CAT_MODELS)],
   });
+  v.extend_from_slice(&binds! {
+    action: Action::LaunchNamed, scopes: FocusSet::LIST,
```
deepu105: Behavior change request, not a bug fix: I want Alt+Enter off the model list entirely.

New flow:

- Model list: Enter opens the launch picker, and that's it. Alt+Enter does nothing here.
- Launch picker: Enter launches unnamed as today, Alt+Enter asks for the name and launches `<model-id>@<name>`.

So `LaunchNamed` moves from `FocusSet::LIST` to `FocusSet::RIGHT_PANE` (the surface `Submit` already lives on), the list-side handler in `events.rs` goes away, and the dialog's accept path submits the launch directly off the picker that's already open, instead of staging `commit_launch_name` into a freshly built picker. That staging path is also what breaks today when the right pane isn't on the Settings tab, so this reshape removes that bug rather than patching it.

Two follow-ons while at it: the dialog hint should show Enter as the accept key (Alt+Enter can't submit it anyway), and D5's "help bar" wording in the plan needs updating, since the hint now belongs in the picker rather than the list chip row.

deepu105 (commented):
One behavior change request on top of the DRY pass: move Alt+Enter off the model list.

deepu105 (commented):
Rest of my review pass that isn't covered by the DRY and behavior comments: an empty `--name` slips through, `list` and `show` collapse concurrent launches of one model, the umbrella route drops the parsed name, the Alt+Enter label misses the platform split, and the docs and plan side of the feature hasn't shipped. Inline below.

src/cli/cli_args.rs
```diff
@@ -408,6 +408,12 @@ pub struct StartArgs {
   /// catalog. Non-interactive callers (CI / piped / `--json`) must
   /// pass an explicit reference.
   pub model: Option<String>,
+  /// User-chosen name for this launch. When set, the launch is
+  /// addressable by `<model-id>@<name>` in `stop`, `logs`, and the
+  /// proxy's `body.model`. A second launch of the same model with the
+  /// same name is refused (the name is unique per model).
+  #[arg(long, value_name = "NAME")]
```
deepu105: `--name` takes an empty or whitespace-only string as-is, and nothing downstream rejects it. `start qwen3 --name ""` passes the daemon's duplicate gate, gets stamped on the snapshot, and then the two ends disagree: `/v1/models` publishes `qwen3@` because the emission loop doesn't guard the suffix, while `decide` rejects the split on the `!n.is_empty()` guard. So the id is listed but unaddressable. The TUI dialog converts empty to `None`, the CLI and the daemon don't. Trim and reject at parse time, or normalize empty to `None` in `build_payload` before it reaches the daemon.

src/cli/output.rs
```diff
@@ -179,7 +179,13 @@ fn running_status_cell(row: Option<&RunningRow>) -> String {
   } else {
     port_part
   };
-  format!("{glyph} {state_label} {port_part}")
+  // A named launch is addressable as `<model-id>@<name>`; surface the name
+  // here so `list` shows which named launch of this model is running.
+  let name_part = match r.name.as_deref() {
```
deepu105: This cell can only ever show one `@name`, because upstream `running_index` (`src/cli/resolve.rs`) keeps one row per path and drops the rest, and `show` matches with `.find()` on top. The feature's own motivating case, `qwen3@coder` and `qwen3@writer` both running, renders as a single catalog row carrying whichever `@name` won, and `show qwen3` reports only one launch. Step 3 also wanted the addressable `<model-id>@<name>` string shown whole, not an `@name` fragment appended to the status cell. The index needs to become path to Vec<row>, and `list`/`show` need a line per running launch.

src/proxy/route.rs
```diff
@@ -249,21 +255,39 @@ pub(crate) async fn decide(state: &Arc<ProxyState>, body_model: Option<String>)
...
+          // the client nothing to refine with. Every entry here routes.
+          return RouteDecision::Ambiguous {
+            requested_model: requested,
+            candidates: crate::launch::resolve::published_ids_for(&rows, &candidates),
+          };
+        }
       };
+      (n, r)
     }
   };
```
deepu105, line 295: The resolve block now parses a name, but the umbrella early-return just below (`decide_umbrella_route(state, requested, &resolved)`) never sees it. A delegated lemonade launch gets its name stamped on the snapshot, `/v1/models` publishes `Qwen2.5-0.5B-Instruct@coder` for it, and a request for that id goes down the umbrella path with the name dropped. The published address is cosmetic and D4's auto-start-with-name can't happen for managed-multiplexer models. Either thread the name into `decide_umbrella_route` and honor it, or stop publishing named ids for umbrella rows.

src/tui/keybindings.rs
```diff
@@ -1398,6 +1408,10 @@ pub const SHIFT_TAB_LABEL: &str = "⇧↹";
 /// both platforms.
 pub const SHIFT_ENTER_LABEL: &str = "⇧⏎";
 
+/// `Alt+Enter` chord label (the `Alt+⏎` "launch as…" quick-jump). Same on
+/// both platforms.
+pub const ALT_ENTER_LABEL: &str = "⌥⏎";
```
deepu105: This one skips the platform split `ALT_PREFIX` does right above it: `⌥` on macOS, `Alt+` everywhere else. On Linux and Windows the help overlay and the dialog hint render the Apple glyph while every other Alt chord renders with the `Alt+` prefix. D5 specified "⌥⏎ on macOS, Alt+⏎ elsewhere". Same cfg split, or compose it from `ALT_PREFIX` plus the enter label so there's one source of truth.

docs/plans/2026-09-03-002-feat-named-launches-plan.md
```diff
@@ -0,0 +1,319 @@
...
+      not one (D4a). This is the single-flight regression that would otherwise be
+      invisible — it looks like the feature working while one session silently
+      shares the other's instance.
+- [ ] `/v1/models` lists the named row while live and drops it after `stop`.
+- [ ] TUI: `Alt+⏎` on the list opens the dialog; `⏎` still opens the picker
+      directly. Golden snapshot covers the help-bar hint (D5), the list pane's
+      `@name` suffix, and the Settings row.
+- [ ] The proxy auto-start path sets `name` but still cannot set `force`.
+
+### Step 6 — docs
```
deepu105, line 321: Step 6 hasn't shipped: the diff touches no docs except this plan. Missing are the CHANGELOG `[Unreleased]` line, the usage.md entries for `start --name`, the `model@name` reference form, and the new `status.name` JSON field, and the architecture.md routing note. TODO.md's R9 entry is still open too. AGENTS.md is explicit that docs ship with the code in the same commit, and the step boxes in this plan are all unticked. From step 5, the golden snapshot coverage (list `@name` suffix, hint, Settings name row) also hasn't landed.

src/cli/resolve.rs
```diff
@@ -490,6 +496,29 @@ pub fn resolve_running(rows: &[RunningRow], reference: &str) -> Result<RunningRo
...
+  // The model part is matched against the path (same as the fallback below);
+  // the name part must match exactly (case-insensitive).
+  if let Some((model_ref, name_ref)) = needle.split_once('@') {
+    let name_lower = name_ref.to_lowercase();
+    let by_named: Vec<&RunningRow> = rows
+      .iter()
+      .filter(|r| {
+        r.name
+          .as_deref()
+          .map(|n| n.to_lowercase() == name_lower)
```
deepu105: The case-insensitive compare here turns the proxy's case-sensitive match into a dead end rather than just an inconsistency. Reproduced on `ca672e5` against a live daemon:

```
# the existing launch is named `coder`
$ curl -s :41970/v1/chat/completions -d '{"model":"Llama-3.2-1B-Instruct-Q4_K_M@CODER",...}'
http=200                                  # auto-started L4 named `CODER`, a second full model load

$ llamastash status
L3  ready  chat  41101  191617   4096  coder
L4  ready  chat  41103  293163  131072  CODER

$ llamastash stop Llama-3.2-1B@coder
✗ `Llama-3.2-1B@coder` matches 2 launches: L3, L4      (exit 66)
```

Once the case variant exists, this line matches both rows, so neither launch is stoppable by name any more. Only the launch id or the port works. The case split is therefore not only a routing inconsistency, it is unrecoverable through the exact surface the feature exists to provide, and on a 27 GB model the auto-start it triggers is expensive before you notice.

Making the canonical predicate case-insensitive on both sides fixes the direction of the failure too: the daemon gate would then reject `CODER` against a live `coder` before the second load happens, instead of the proxy treating them as two addresses.

src/cli/resolve.rs
```diff
@@ -490,6 +496,29 @@ pub fn resolve_running(rows: &[RunningRow], reference: &str) -> Result<RunningRo
...
+          .unwrap_or(false)
+          && {
+            let path = std::path::Path::new(&r.model_path);
+            let fname = path.file_name().and_then(|s| s.to_str()).unwrap_or("");
+            let parent = path.parent().and_then(|p| p.to_str()).unwrap_or("");
+            fname.to_lowercase().contains(&model_ref.to_lowercase())
```
deepu105, line 543: An empty model half makes this match every row: `"".contains("")` is true, so `stop @coder` silently becomes a cross-model name lookup and fails with an ambiguous error listing launches of unrelated models. The same line covers the parent-directory check below it.

The proxy guards the name half with `!n.is_empty()` in `decide`; this path guards neither half, so `stop qwen3@` also matches an empty-named launch, which per the `--name ""` comment can exist. An `!model_ref.trim().is_empty()` guard next to a name guard would make both spellings error instead of widening the match set.

src/cli/output.rs
```diff
@@ -515,7 +525,10 @@ pub fn status_human(snap: &StatusSnapshot) -> String {
         Some(c) => c.to_string(),
         None => "-".to_string(),
       };
-      let name = r.name();
+      // A user-chosen launch name (from `--name`) is the distinguishing
+      // identifier for a named launch; show it in the NAME column when set,
```
deepu105: Separate from the `list` status cell above: `status` has no model column, so replacing the display name here drops model identity from the row entirely.

```
LAUNCH_ID  STATE  MODE  PORT    PID     CTX    NAME
L2         ready  chat  41102   183378  4096   reviewer
L3         ready  chat  41101   191617  4096   coder
```

Two different models both named `coder` are indistinguishable here, and `status` is the command you reach for to work out what to stop. `status_json` is unaffected, it keeps `model_path` and adds `name` beside it, so this is the human table only. Either keep the model name and append the launch name, or add a MODEL column and let NAME hold the launch name.

deepu105 (commented):
Additions to my earlier passes. Everything below is new relative to the DRY, behavior, and docs comments already on this PR; the case-variant one extends the `launch_service.rs:331` thread.

**The PR body describes commands that do not work.** Verified against a live daemon on `ca672e5`, isolated state dir, one model in the catalog, two launches named `coder` and `reviewer`:

```
$ llamastash stop coder
✗ no running launch matches `coder`                     (exit 66)
$ llamastash show coder
                                                        (exit 66)
$ llamastash show Llama-3.2-1B@coder
✗ no model matches `Llama-3.2-1B@coder` (1 known)       (exit 66)
$ llamastash stop Llama-3.2-1B@coder
✓ stopped L1 -> stopped                                 (exit 0)
```

There is no bare-name resolution anywhere in `resolve_running`, and `show` resolves against the catalog only, so it never accepts the `@name` form even though the body says "name-aware stop/show/list/resolve". Either land D3's bare-name convenience or narrow the body to what ships: `stop <model>@<name>`, `logs <model>@<name>`, and name-aware output.

**`build_log_path` collides for two launches of one model (`src/daemon/launch_service.rs:1793`).** Pre-existing, not introduced here, but named launches make it a normal path instead of an odd one. The filename is `{stem}-{blake3[0..8]}-{unix_secs}.log` with no launch id, so two launches of the same model started in the same wall-clock second open the same file. I hit it twice in about five attempts:

```
Llama-3.2-1B-Instruct-Q4_K_M-c3a8a340-1788583275.log: listening on :41101  listening on :41102
Llama-3.2-1B-Instruct-Q4_K_M-c3a8a340-1788632515.log: listening on :41101  listening on :41102
```

When it does not collide the two tails are correctly separate, so this is timing-dependent rather than always-on. It matters because `logs <model>@<name>` is the per-launch diagnosis surface, and on collision it returns both processes interleaved. Adding the launch id to the filename is a one-line fix and would have been invisible before this feature.

**`start` never reports the name it just set (`src/cli/start.rs:775`).** The headline uses `row.name()`, the model name:

```
✓ started Llama-3.2-1B-Instruct-Q4_K_M.gguf -> launch_id=L1 port=41101 pid=353644
```

Step 3 asks for the name in the success line. Worth including since the name is the thing the user has to remember to address the launch later, and it is the only confirmation that the daemon accepted it rather than silently dropping it.

**Three small ones.** `LaunchNameDialog::error` is never set to `Some` anywhere, so the field, the spacer row, and the render branch at `launch_name_dialog.rs:114` are dead. `commit_launch_name` takes a `writer` it discards with `let _ = writer;` (`events.rs:1358`). And the duplicate-name refusal names the name but not the launch that holds it, so the user has to run `status` to find what to stop; the launch id belongs in that message.

**What I confirmed working**, so it is on the record: two named launches of one model land on distinct ports; the proxy routes `@coder` and `@reviewer` to their own supervisors; `stop` and `logs` both accept `<model>@<name>`; `/v1/models` publishes `Llama-3.2-1B-Instruct-Q4_K_M@coder` beside the bare id; D4 auto-start carries the name (stopping `coder` then requesting `@coder` spawned a fresh launch named `coder`); names persist to `state.json`; the orphan sweep clones the whole `RunningSnapshot` at `orphans.rs:191`, so names survive re-adoption. `cargo test --features test-fixtures` is 2345 lib tests plus all integration binaries green, and `make lint` is clean.

src/daemon/launch_service.rs
deepu105: One more on the same gate: the check and the insert are not atomic, so a shared predicate alone would still let two concurrent starts through.

`ctx.state.snapshot()` (`src/daemon/context.rs:150`) takes the mutex, clones, and releases it here. The row is not pushed until `spawn_supervised` (~line 1267), after the process is spawned and the supervisor is registered. Nothing holds a lock across check to insert, and there is no launch-level mutex in the daemon.

Worth noting the port allocator does serialize (`registry.rs:48`, `reserved_ports: Arc<TokioMutex<BTreeSet<u16>>>`), so the two racing launches land on clean distinct ports and the only symptom is two rows carrying the same name. That is exactly the ambiguity the gate exists to prevent, and it arrives with no error at all.

So the canonical predicate fixes the case axis but not this one. The gate needs to run in the same critical section as the insert, or the name needs reserving alongside the port.

docs/plans/2026-09-03-002-feat-named-launches-plan.md
deepu105: Two more Step 5 items on top of the golden snapshots:

- The `state.json` round-trip byte-identical test for an unnamed row, which the plan marks not optional. `RunningSnapshot` gained a field at `state_store.rs:153` and nothing pins that an unnamed row still serializes to the pre-feature bytes.
- No test that a name survives a daemon restart through orphan re-adoption. The mechanism is there and reads correctly (`orphans.rs:191`, `adopted.push(snap)` clones the whole snapshot so `name` rides along), but D1's whole justification for putting the name on the snapshot instead of only on the supervisor is restart survival, and that is the one load-bearing path with no coverage.

src/proxy/route.rs
deepu105: Fixed in `2e2e2ae`. There is one split now: `parse_named_reference` in `src/launch/resolve.rs`, called by `route::decide`, `cli::resolve::resolve_running` and `show`. It splits on the last `@` per D2, requires both halves non-empty, and each caller keeps the whole-string-first fail-safe so a `foo@bar.gguf` file still resolves as a plain reference.

The comparison moved out with it. `name_matches` is the only place case is decided, and it is case-insensitive on every surface, so the three axes you found cannot drift apart again.

src/proxy/router.rs
deepu105: Fixed in `126753d`. `named_launch_ids` is the one producer; `list_models` and `ollama_tags` dress the same pairs in their own row type.

The model half now comes out of the same `published_ids` index the catalog rows use, so two same-basename GGUFs in different roots publish the disambiguated stem instead of a bare `qwen3@coder` that then 400s on the split. D6 holds: `published_id_index` is still the single rule for the model half. Two unit tests pin it, one that the published id is taken rather than the stem, one for the off-catalog fallback plus the dedup.

src/proxy/route.rs
deepu105: Fixed in `2e2e2ae`. `RunningSnapshot::carries_name` is the single predicate, called by `route::decide`, `attach_target`, the daemon gate and the CLI resolver.

It keys on `launch_id` as you suggested. The port stays only as the fallback for an adopted row whose `launch_id` is `None`, which is the same rule `drop_running_snapshots` already uses.

src/daemon/launch_service.rs
deepu105: Fixed in `2e2e2ae`, all four parts.

**One predicate.** `RunningSnapshot::carries_name`, keyed on `launch_id`, comparing through `name_matches`, so the gate, both proxy walks and the CLI resolver cannot disagree.

**Case.** `eq_ignore_ascii_case` everywhere. Verified on the rebased branch against a live daemon and a real `llama-server`: `start <model> --name CODER` with `coder` live is refused with ``name `CODER` is already running as L1``, and a proxy request for `<id>@CODER` attaches to the existing launch. The per-launch logs confirm only two upstreams ever existed, so the second full load is gone.

**Error state.** The gate skips rows whose supervisor has errored, so a launch stuck in `error` no longer holds its name until someone stops it by hand.

**Atomicity (your follow-up).** You were right that the shared predicate alone does not close it. The name is now claimed for the duration of the spawn through a `Drop` guard over a registry set, mirroring `reserved_ports`, so the check and the insert sit in one critical section from the caller's point of view. Two concurrent `start --name coder` now yield one launch and one ``name `coder` is already being started for this model``, instead of two rows sharing an address with no error at all.

src/proxy/coalesce.rs
deepu105: Fixed in `bdc5313`. `FlightKey` names the key in `Leader::key`, `acquire` and the test helper, so the shape has one name.

src/daemon/state_store.rs
deepu105: Fixed in `126753d`. `test_support::running_row` is a builder with every field pre-filled and one setter per field a test actually varies. All 15 literals across the 8 inline modules and the 4 integration suites go through it now, so the next field added to `RunningSnapshot` is one default in one place.

src/tui/launch_name_dialog.rs
deepu105: Hint fixed in `020538d`, extraction deferred as you suggested.

The lookup now resolves each action under the focus it is actually scoped to (`Submit` under `RightPane`, `Cancel` under `ConfirmPopup`), and it shows the Enter label, since the field submits on a bare Enter and `Alt+Enter` cannot accept it anyway. So a `keybindings:` override does show up here now.

The `error` field is no longer dead either: a blank-but-typed name is refused inline, matching what `--name` does. Extraction trigger recorded as a `TODO.md` line, to be pulled out when a third single-field modal appears.

src/tui/keybindings.rs
deepu105: Done in `020538d`, reshaped the way you asked.

`LaunchNamed` is `FocusSet::RIGHT_PANE` now, the list-side handler is gone, and the dialog's accept path stamps the name onto the picker that is already open and submits it. The staging path does not exist any more, so the Logs/Chat/Embed/Rerank bug is removed rather than patched.

Both follow-ons are in: the hint shows Enter as the accept key, and D5's "help bar" wording in the plan now says the hint belongs to the picker footer. The footer advertises `Alt+⏎:launch as…` at rank 11, right behind `⏎:launch`, so it survives width pressure.

Two new goldens cover it: `launch-picker.txt` for the footer chip, `named-launch.txt` for the list row's `@coder` suffix together with the Settings `name` row. Smoke tests assert Alt+Enter does nothing on the list, opens the dialog on the picker, and that accepting dispatches a `start_model` carrying the name.

src/cli/cli_args.rs
deepu105: Fixed in `2e2e2ae`. `--name` is trimmed at parse time and a whitespace-only value is a usage error rather than a silently dropped flag.

The TUI dialog matches now: nothing typed still launches unnamed, but a blank-but-typed name is refused inline instead of being quietly converted to `None`. Since no empty name reaches the daemon, the emission loop and `decide` can no longer disagree about `qwen3@`.

src/cli/output.rs
deepu105: Fixed in `16fa9fb`. `running_index` returns `HashMap<String, Vec<RunningRow>>`, `list` prints one line per live launch, and `show` prints one block per launch (its `running` key is an array now, `[]` rather than `null` when nothing is live).

The address moved into the identity column as Step 3 specified: NAME reads `qwen.gguf@coder`, the whole string a user pastes into a client, and STATUS is back to what it was. `list --json` gained a `launches` array so an agent sees the same rows the table prints; `status` stays the first launch, since agents pin `models[i].status.state`.

Also from your review body, since it lands in the same place: `stop coder`, `show coder` and `show <model>@coder` work now rather than exiting 66. `resolve_running` gained the unique-bare-name tier ahead of the path walk, the way an exact launch id already ranked, and `show` resolves through three tiers (catalog, then `<model>@<name>`, then the live launches), so `show L3` and `show 41100` reach the launch's model too.

src/proxy/route.rs
deepu105: Fixed in `126753d`, but not either option verbatim, so worth your eye.

The daemon now refuses `--name` on a managed-multiplexer model (`multiplexer_refuses_name`). Such a backend serves every one of its models from one shared process, so a name there was never going to select an instance. Filtering the emission alone would have left `status` showing a name that addresses nothing, which is the same lie one level down.

With no delegated row able to carry a name, the umbrella early-return has nothing to honor and nothing unroutable is published. The early-return carries a comment saying so, and a stale id from before still reaches the model it asks for.

Happy to switch to threading the name through `decide_umbrella_route` instead if you would rather keep `--name` accepted for those models.

src/tui/keybindings.rs
deepu105: Fixed in `020538d`. `ALT_ENTER_LABEL` is `crate::alt_label!("⏎")` now, the same macro every other Alt chord goes through, so the platform split has one source of truth rather than a second hardcoded glyph.

Confirmed in the new `launch-picker.txt` golden, which is captured on Linux and renders `Alt+⏎:launch as…`.

docs/plans/2026-09-03-002-feat-named-launches-plan.md
deepu105: Shipped: docs in `fc5330c`, the missing tests in `bdc5313` and `020538d`.

**Docs.** `usage.md` covers `start --name`, the `model@name` reference form on `stop` / `logs` / `show`, the `name` field on `status` rows, the `launches` array on `list --json`, the named ids the proxy publishes and where they come from, the picker keybinding, and the manual pi / opencode step from D7. `architecture.md` gained a Named launches section stating the four rules in one place: where the split happens, how names compare, what the namespace covers, and why the name lives on the running snapshot. `CHANGELOG.md` has its line under `[Unreleased]`. `TODO.md`'s R9 proxy-ambiguity entry is closed, the D7 patcher entry stays open as high priority, and the shared-modal-frame note is filed.

**Tests from your follow-up.** `an_unnamed_running_row_is_byte_identical_to_a_pre_name_row` pins that an unnamed row emits no `name` key at all, not `"name": null`, and that a named one round-trips. `an_adopted_row_keeps_its_launch_name` drives the orphan sweep against a live pid and port and asserts the name survives re-adoption, which is D1's whole justification.

Golden coverage is the two fixtures described in the Alt+Enter thread. Every step box in the plan is ticked, and the plan carries a Review section with all 36 findings and what each one became.

src/cli/resolve.rs
deepu105: Fixed in `2e2e2ae`, and your reading of the direction was right: making the shared predicate case-insensitive fixes it at the daemon rather than papering over it at the proxy.

Reproduced your scenario on the rebased branch, isolated daemon, real `llama-server`, `Qwen3.5-4B-Q4_K_M`:

```
$ llamastash start Qwen3.5-4B --name CODER          # `coder` already live as L1
✗ name `CODER` is already running as L1              # refused before any load

$ curl -s :41935/v1/chat/completions -d '{"model":"Qwen3.5-4B-Q4_K_M@CODER",...}'
http=200                                             # served by L1, no second launch

$ llamastash status
LAUNCH_ID  STATE  MODE  PORT   PID     CTX   NAME
L1         ready  chat  41100  192454  4096  Qwen3.5-4B-Q4_K_M.gguf@coder
L2         ready  chat  41101  193104  8192  Qwen3.5-4B-Q4_K_M.gguf@writer
```

Two launches, not three. The per-launch logs confirm the `@CODER` requests landed on L1's upstream. `stop <model>@coder` resolves to exactly one launch, so the unrecoverable state is unreachable now.

src/cli/resolve.rs
deepu105: Fixed in `2e2e2ae`. `parse_named_reference` requires both halves non-empty, so `@coder` and `qwen3@` are not name references at all and fall through to the plain reference path instead of widening the match set.

Live check: `stop @coder` exits with ``no running launch matches `@coder` `` rather than an ambiguity listing unrelated models. `resolve_running_does_not_widen_on_an_empty_half` covers both spellings.

src/cli/output.rs
deepu105: Fixed in `16fa9fb`. The NAME cell keeps the model and appends the launch name, so it reads `qwen.gguf@coder`, which is also the exact string `stop` takes. Same join `list` uses in its NAME column.

Two different models both named `coder` are distinguishable again, pinned by `status_human_keeps_the_model_next_to_the_launch_name`. `status_json` is untouched.

src/cli/start.rs
```diff
@@ -31,6 +31,19 @@ use crate::ipc::Client;
...
+  if args
+    .name
+    .as_deref()
+    .is_some_and(|raw| raw.trim().is_empty())
+  {
+    return Err(CliExit::new(
+      USAGE,
+      "`--name` needs a non-empty name (letters, digits, `-` or `_`)",
+    ));
+  }
```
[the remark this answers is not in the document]
deepu105: Good catch, and the messaging was the smaller half of it. Fixed in `b153327e`, extended in `96d9ae3`.

`validate_launch_name` (trimmed, non-empty, ASCII letters/digits/`-`/`_`) now runs at the flag's value parser, in the TUI dialog, and in the daemon's `start_model` gate for raw JSON-RPC callers, so no entry point can stamp a name that does not round-trip through the split.

You were right that the `@` case was the dangerous one, and there was a second half I missed on the first pass: the reader still accepted anything non-empty. So `qwen3@co der` parsed as a named reference, missed, and auto-started, which the daemon then refused. Each attempt burned one of that model's three auto-start failures per minute, and that budget is keyed on the model id, so three typos would have suppressed auto-start for every name of that model. Reader and writer now share one `is_launch_name` predicate, and a malformed address is a plain model-not-found.

Malformed values exit 64 (`USAGE`), same as every other bad flag value on this binary.

src/daemon/registry.rs
```diff
@@ -180,6 +220,21 @@ impl SupervisorRegistry {
...
+  reserved: Arc<StdMutex<BTreeSet<(String, String)>>>,
+  key: (String, String),
+}
+
+impl Drop for NameReservation {
+  fn drop(&mut self) {
+    if let Ok(mut reserved) = self.reserved.lock() {
+      reserved.remove(&self.key);
+    }
+  }
```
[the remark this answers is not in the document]
deepu105: Correct, and it would have held the name until the daemon restarted. Fixed in `37b82a6a`: `Drop` recovers the guard the same way `lock_names` does (`unwrap_or_else(|p| p.into_inner())`), since the set holds plain keys and there is nothing to be poisoned about.

`a_poisoned_lock_does_not_leak_a_held_name` poisons the mutex under `catch_unwind`, drops the claim, and asserts the name is claimable again.

src/tui/events.rs
```diff
@@ -1289,6 +1306,64 @@ fn commit_save_preset(app: &mut App, writer: Option<&mpsc::Sender<WriterCmd>>) {
...
+    InputOutcome::Submit => {
+      // Nothing typed launches unnamed; whitespace is a typo, not a name, and
+      // the CLI rejects `--name "  "` for the same reason.
+      if !dialog.input.buffer().is_empty() && dialog.name().is_empty() {
+        dialog.error = Some("a name cannot be only whitespace".to_string());
+        return;
+      }
+      dialog.error = None;
+      // Borrow released below; the picker launches carrying the typed name.
+    }
```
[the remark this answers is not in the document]
deepu105: Yes, same rule on both ends now. Fixed in `b153327e`: the dialog runs the same `validate_launch_name` the `--name` parser does and refuses inline, so nothing typed there can publish an address that will not parse back. Nothing typed at all still launches unnamed.

`a_name_the_cli_would_refuse_is_refused_inline_too` drives `a@b` through the dialog and asserts it stays open with the error.

src/tui/list_pane.rs
```diff
@@ -1174,7 +1191,17 @@ fn render_row<'a>(
...
       spans.push(marker_span(*state, *favorite, palette));
-      spans.push(Span::raw(cell(name.as_str(), name_w)));
+      // A named launch renders `<name>@<launch-name>` so the row reads the
+      // same addressable string the `list` table and proxy use. The `@name`
+      // suffix is muted so the model name stays the dominant token.
+      match launch_name {
+        Some(n) if !n.is_empty() => {
+          let mut label = name.clone();
+          label.push_str(&format!("@{n}"));
+          spans.push(Span::raw(cell(&label, name_w)));
```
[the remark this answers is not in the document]
deepu105: Fixed in `90ecbc0c`: `String::with_capacity(1 + n.len())` plus `push`/`push_str`, so the temporary is gone and the name is no longer cloned either.

Worth noting the follow-up, since it came out of the same line. Splitting the suffix into its own span to mute it broke the selected row: `highlight_style` REVERSEs each cell against whatever fg it carries, so the muted suffix inverted to a grey block mid-strip while the cells around it flipped to the row colour. Golden snapshots are text-only, so nothing caught it. `96d9ae3` mutes the suffix only while the row is unselected, and three tests now read real buffer cells rather than spans.

tests/proxy_fallback.rs
```diff
@@ -278,12 +287,26 @@ fn header_value<'a>(headers: &'a [(String, String)], name: &str) -> Option<&'a s
...
+  // Mirror the production `drop_running_snapshots` so the state
+  // snapshot reflects the stopped launches (needed for /v1/models).
+  if !stopped_ports.is_empty() {
+    let ports = stopped_ports.clone();
+    ctx
+      .state
+      .mutate(move |s| {
+        s.running.retain(|r| !ports.contains(&r.port));
+      })
+      .await;
```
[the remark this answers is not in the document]
deepu105: Fixed in `89354643`, `HashSet` it is. The clone per row went with it, since the set now moves into the closure.

deepu105:
Approved

Issue #36, opened 2026-06-15
Feature request: SGLang backend support


Ansonfishing:
LlamaStash is a great TUI launcher for llama.cpp GGUF models. Would it be possible to add support for **vLLM** and **SGLang** as additional backends?

**Use case:**
Many users (myself included) run multiple inference engines side by side:
- **llama.cpp** for GGUF models (already supported)
- **vLLM** for safetensors/AutoRound/GPTQ/AWQ/FP8 models
- **SGLang** for structured output and high-throughput serving

Currently all three engines require separate management workflows. A unified TUI like LlamaStash could auto-detect installed runtimes (vLLM, SGLang) alongside llama.cpp, scan safetensors model directories, and let users launch any backend with one keystroke.

**Proposed behavior:**
1. During `init`, detect `vllm` and `sglang` binaries/venvs
2. Scan model directories for safetensors model dirs (config.json + *.safetensors)
3. Allow per-model backend selection (llama.cpp / vLLM / SGLang)
4. Save named launch profiles per model with engine-specific args (--gpu-memory-utilization, --mem-fraction-static, --port, etc.)

**Alternative approaches considered:**
- [llml](https://github.com/flyingnobita/llml) — TUI launcher that supports llama.cpp + vLLM, but not SGLang yet
- Manual CLI scripts — works but lacks the unified TUX experience

Happy to help test if this is on the roadmap!

deepu105:
vLLM is already supported via the Lemonade backend (https://github.com/llamastash/llamastash/blob/main/FEATURES.md#npu--multi-engine-via-lemonade-experimental-opt-in) would this work for you or do you think there is more value in supporting vLLM directly. I'll look into adding SGLang. 

Btw if you are interested in giving it a shot, you could try (with LLM or not) to add an SGLang backend. I made sure backends are easily pluggable in architecture.

deepu105:
vLLM support landed in v0.2.0

PR #78, opened 2026-09-09
fix(unit5): catch a backend declaring one knob id twice

bernardogv:
`validate()` keys its `kinds` map by id across all backends, so a repeat whose kind matches falls through the `Some(_) => {}` arm. Sharing an id across backends is deliberate, so that arm has to stay — but one backend declaring the same id twice reached no check at all, and `duplicate_flags` misses it because two knobs can share an id while emitting different flags. Every lookup resolves to the first match, so the second declaration was silently unreachable: its ring, concept and emit never applied.

This adds `duplicate_ids`, mirroring `duplicate_flags`, in the same per-backend loop, with a `DuplicateId` error and two tests. The main test asserts `duplicate_flags` finds nothing on the same input, so it can only pass because of the new check.

No shipped backend trips it today (`registry_is_valid` passed before and after). It is relevant now because the SGLang knob table (#36) is carrying a local `knob_ids_are_unique` test to cover exactly this gap; once SGLang is registered in `Backends::all()` that local test can go.

Internal validator only — no user-visible behaviour, so no CHANGELOG entry; happy to add one if you'd prefer.

`make test`, `make lint`, knob-parity all green; the change was also mutation-checked (disabling `duplicate_ids` fails `one_backend_cannot_declare_an_id_twice`).

### DIFF
--- a/src/launch/knobs/registry.rs
+++ b/src/launch/knobs/registry.rs
@@ -246,6 +246,14 @@ pub enum RegistryError {
     first: &'static str,
     second: &'static str,
   },
+  /// One backend declares the same id twice. Every id lookup resolves to the
+  /// first match, so the second declaration is unreachable: its ring, concept
+  /// and emit never apply. `KindConflict` only sees a repeat that changes kind,
+  /// and sharing an id *across* backends is deliberate, so neither covers this.
+  DuplicateId {
+    backend: &'static str,
+    id: &'static str,
+  },
 }
 
 impl std::fmt::Display for RegistryError {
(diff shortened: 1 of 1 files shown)


bernardogv:
The three red jobs are an install-step flake, not this diff.

`shellcheck install.sh`, `bats install.sh (ubuntu-latest)` and `cross-compile aarch64-linux` all fail inside their `apt-get install` step with a `dl.google.com` `chrome-stable` `Packages.gz` **Hash Sum mismatch** (exit 100) — the runner image's pre-baked Chrome apt repo caught mid-publish. The tell is that each one dies in an *install* step within seconds, before touching any code.

Every job that actually compiles or tests is green across all three OSes, and those same three jobs pass on `main` at `c4ab821`, which is this branch's base (ahead by 1, behind by 0).

A re-run should clear it, but I'm pull-only here so I can't trigger one — happy for you to kick it whenever convenient.

deepu105:
Thanks. The knob/preset layer was reworked recently, so I'll review and rebase this on top of that.

PR #79, opened 2026-09-09
feat(sglang): SGLang backend for safetensors repos (#36, second half)

bernardogv:
Second half of #36: SGLang as a native process-per-model backend for safetensors HF repos, on the substrate the vLLM backend (#63) established.

## What lands

- **`src/backend/sglang/`** — knob table, `Backend` impl, and a unified-memory guard that differs from vLLM's on purpose. SGLang has no byte-level KV cap: `--mem-fraction-static` is a fraction of the *whole* pool (system RAM on a DGX Spark or Strix Halo), and the only deterministic bound is `--max-total-tokens`, in tokens. The guard takes the same byte budget vLLM gets and divides it by the model's KV bytes per token, read from `config.json` (GQA/MHA, multi-head latent attention, and `text_config` shapes). Unreadable geometry **refuses** the launch, naming `max-total-tokens` as the override, rather than guessing.
- **Registration**: enum variant, `for_each_backend!` arm, `Backends::all()`, `BackendConfig.sglang`, config re-export, `--sglang` / `LLAMASTASH_SGLANG=1`.
- **Fixture + integration tests** (`tests/fixtures/fake_sglang_server.rs`, `tests/sglang_backend_test.rs`): discovery under a configured root, the merged row when both engines are on, a launch whose token cap is asserted on the argv the fixture was handed, the refusal without geometry lifted by an explicit cap, early-bind readiness, orphan re-adoption, force flag, and the no-leak guard.
- **Docs**: `docs/sglang-setup.md`, `docs/plans/2026-09-09-001-…`, `scripts/sglang/{probe.sh,uat.sh}`, plus README / CHANGELOG / usage / architecture / troubleshooting / `config.example.yaml` / TODO / AGENTS.md drift.

## Generic changes this needed (each its own commit)

- `refactor(unit4)`: the safetensors leaf helpers (`eligible`, `project`, `is_safetensors_snapshot`, `repo_id_for_snapshot`, `weights_bytes`) move from `backend/vllm/discovery.rs` into `discovery::hf_repos`, taking the backend id as a parameter. The vLLM leaf file is deleted; tests moved with the code.
- `refactor(unit5)`: the unified-host cache budget (`DEFAULT_KV_CACHE_BYTES`, `MIN_KV_CACHE_BYTES`, `UNIFIED_HOST_RESERVE_BYTES`, the refuse-not-floor arithmetic) moves to `launch::admission`; vLLM calls it.
- `refactor(unit5)`: the extras strip moves to `launch::params::strip_forbidden_extras`.
- `fix(unit4)`: **two safetensors engines projecting one snapshot used to leave one row** — the catalog is keyed by path, so the last projector's row silently replaced the first and the repo listed a single backend. `discovery_task::merge_by_path` folds them into one row, highest launch priority first, so `supported_backends` reads `["vllm", "sglang"]`.

## Decisions worth a look

- `served-model-name` is deliberately **not** a knob (the launcher owns it; argparse keeps the last occurrence, so a user-set one would leave readiness waiting out its budget) and is refused in extras.
- `quantization` takes the vLLM declaration verbatim because the registry allows one kind per id; SGLang's dashed methods (`auto-round`) go through extras. `tool-call-parser` / `reasoning-parser` are closed `Enum`s from 0.5.18's own argparse choices.
- Launch priority 4, below vLLM's 5: with both installed, `auto` keeps picking vLLM; `--backend sglang` selects SGLang.
- `--enable-unified-memory` is exposed but is **not** a host-UMA switch — per 0.5.18's help it changes the pool layout for hybrid (SWA/Mamba) models. The setup doc says so.

## Verified against

SGLang `0.5.18` (`lmsysorg/sglang:v0.5.18-cu130`) on a DGX Spark: `sglang serve --help` for the flag surface and choice lists, and a live server's `/get_server_info` / `/v1/models` for what the actuals read. **Load-tested on spark2 (2026-09-09)** via `scripts/sglang/uat.sh`: the guard passed `--max-total-tokens 699050` and the server resolved exactly that pool, host memory peaked at 22 GiB of 121 GiB, `resolved_ctx` read back, chat served, replay re-resolved the cap, a preset cap beat it, stop reaped the container. Record: spark-casebook `2026-09-09-llamastash-sglang-token-cap-uat.md`. Two gaps filed in TODO: `resolved_ctx` is empty without `--ctx`, and a host with only a Python engine cannot launch (pre-existing, affects vLLM too). `make test` and `make lint` are green; an isolated daemon shows the backend row in `status --json`, `doctor` and the TUI.

Once #78 merges, `knob_ids_are_unique` in `sglang/knobs.rs` becomes redundant with the registry validator and can go.


deepu105:
Thank you for this. I couldnt work on SGLang on Strix halo since it wasnt supported. Will review and merge. Will appreciate another round of smoke tests when I release next version today or tomorrow.

src/backend/sglang/mod.rs
```diff
@@ -0,0 +1,937 @@
...
+        }
+      }
+      None => {
+        log::warn!(
+          "sglang: no host memory reading yet — capping the KV pool at the default \
+           budget rather than letting the launcher size it against the whole pool"
+        );
+        // Not through the cap arithmetic: that would refuse on the (unknown)
+        // free figure, and the default budget is the point of this branch.
+        u32::try_from(crate::launch::admission::DEFAULT_KV_CACHE_BYTES / per_token.max(1))
```
deepu105: This branch skips the token floor that the sampled path enforces. `max_total_tokens_cap` refuses below `MIN_POOL_TOKENS`, but here we just divide and launch with whatever falls out, so a model costing more than about 4 MiB per token gets a sub-2048-token pool with no refusal and no warning. That's precisely the case the floor exists for, and it's reachable on the first launch after daemon start, before metrics have sampled.

Can you run this through the same floor, or refuse the same way?

src/backend/sglang/guard.rs
```diff
@@ -0,0 +1,226 @@
...
+    config
+  } else {
+    config.get("text_config")?
+  };
+  let layers = field(cfg, "num_hidden_layers")?;
+  let dtype = ["torch_dtype", "dtype"]
+    .iter()
+    .find_map(|k| cfg.get(k).or_else(|| config.get(k)))
+    .and_then(|v| v.as_str());
+  let per_layer = match field(cfg, "kv_lora_rank") {
```
deepu105: This branch triggers on the *presence* of `kv_lora_rank`, before heads are looked at. A config carrying `"kv_lora_rank": 0` as a "not MLA" marker takes the latent path, `per_layer` comes out 0, and the function returns `Some(0)`.

Then the `.max(1)` in `max_total_tokens_cap` turns that into `budget` tokens, clamped to `u32::MAX`, so we hand SGLang `--max-total-tokens 4294967295`. The guard is off on exactly the unified-memory host it exists to protect. Same story if `num_hidden_layers` is 0.

I'd treat a zero per-token cost as unreadable geometry and send it down the refusal path with the rest. Gating the branch on `rank > 0` would cover the likelier half of it.

src/backend/sglang/guard.rs
```diff
@@ -0,0 +1,226 @@
...
+    );
+    let just_fits = MIN_KV_CACHE_BYTES / MIN_POOL_TOKENS;
+    assert_eq!(
+      max_total_tokens_cap(8 * GB + MIN_KV_CACHE_BYTES, 0, just_fits),
+      Some(MIN_POOL_TOKENS as u32)
+    );
+  }
+
+  #[test]
+  fn a_zero_per_token_cost_does_not_divide_by_zero() {
```
deepu105: This pins the behaviour from my other comment. A zero cost currently yields an effectively unbounded cap and the test asserts that's fine. Worth flipping to assert `None` once the zero case refuses.

src/backend/sglang/mod.rs
```diff
@@ -0,0 +1,937 @@
...
+    // the launcher is handed, times what one token costs this model.
+    if let Some(tokens) = knob_u64(params, "max-total-tokens") {
+      if let Some(per_token) = guard::kv_bytes_per_token(&params.model_path) {
+        return Some(tokens.saturating_mul(per_token));
+      }
+    }
+    // A fraction is of the whole pool, not of what is free, and SGLang fills
+    // whatever it is given. Project the same way SGLang spends it so the gate
+    // sees the real demand rather than nothing at all.
+    if let Some(frac) = knob_f64(params, "mem-fraction-static") {
```
deepu105, line 293: Two problems here, and the comment right above says the first one out loud. We multiply by `free_bytes`, not the pool, while the comment says a fraction is of the whole pool.

The second one is worse. I checked 0.5.18's `ServerArgs` and `mem_fraction_static` is "the fraction of the memory used for static allocation (model weights **and** KV cache memory pool)", so it already covers the weights. The gate at `launch_service.rs:1137` then adds `resident_weight_bytes` on top of whatever we return here, so the weights get counted twice.

It's over-refusal so the direction is safe, and vLLM carries the identical copy, which is presumably where it came from. Given you've already hoisted the byte budget and the extras strip, this feels like a third thing to share and fix once. Fine as a follow-up if you'd rather keep this PR focused, but at minimum the comment should stop claiming something the code doesn't do.

src/backend/sglang/mod.rs
```diff
@@ -0,0 +1,937 @@
...
+  let resp = client
+    .get(format!("http://127.0.0.1:{port}/get_server_info"))
+    .send()
+    .await
+    .ok()?;
+  if resp.status().as_u16() != 200 {
+    return None;
+  }
+  const MAX_BODY: usize = 256 * 1024;
+  let mut body = resp.bytes().await.ok()?.to_vec();
```
deepu105: `resp.bytes()` buffers the whole body before we truncate, so the 256 KiB cap protects no memory at all. And if a body ever does go over, truncating leaves invalid JSON and `from_slice` fails, so we get no actuals rather than partial ones.

`orphans::fetch_models_body` has the same shape so you inherited this, but this copy raises the cap 8x for a body you describe as 479 keys. Either stream with a real limit and reject past it, or drop the truncate. Either way the doc comment above shouldn't say the body is bounded, because it isn't.

src/backend/sglang/knobs.rs
```diff
@@ -0,0 +1,338 @@
...
+    fallback: crate::launch::params::LayerLabel::ServerDefault,
+    emit: Emit::BareFlagWhenTrue,
+    ring: Ring::None,
+    volatile: false,
+  },
+  KnobDef {
+    // Closed sets: both parsers are argparse `choices` in 0.5.18, so a value
+    // outside the list fails at spawn anyway. `auto` detects from the chat
+    // template and is the useful default when the model needs a parser.
+    id: "tool-call-parser",
```
deepu105, line 145: I diffed this against `FunctionCallParser.ToolCallParserEnum` on the v0.5.18 tag and it matches exactly, all 34 in order. Same for the reasoning list against `ReasoningParser.DetectorMap`, all 26. Nice transcription.

The catch is v0.5.19 landed on 2026-09-05, before this branch, and its breaking changes rename a parser (`spark` to `spark25`). These are closed `Enum`s, so a user on today's SGLang can be refused a name their server would take. They can still reach it through extras since `--tool-call-parser` isn't in the denylist, so it degrades rather than breaks.

Can you re-verify both lists against 0.5.19 and record the version they were transcribed from in `docs/sglang-setup.md`? Otherwise the next person has no way to tell how stale they are.

src/cli/daemon.rs
```diff
@@ -558,6 +562,7 @@ pub(crate) fn build_options(
   lemonade_cli: bool,
   ds4_cli: bool,
   vllm_cli: bool,
+  sglang_cli: bool,
```
deepu105, line 565: This is now ten positional bools, and the call sites read like this:

```rust
None, None, false, false, None, false, false, false, false, false, &cli, &config
```

Adding the eleventh backend flag is a coin flip. Can you pull them into a small struct while you're in here? Roughly:

```rust
#[derive(Default)]
pub(crate) struct BuildOptionsArgs<'a> {
  pub state_dir: Option<PathBuf>,
  pub proxy_port: Option<u16>,
  pub proxy_host: Option<IpAddr>,
  pub ollama_compat: bool,
  pub no_proxy_fallback: bool,
  pub insecure_no_auth: bool,
  pub backend_force: BTreeMap<String, bool>,
  pub cli: &'a Cli,
  pub config: &'a Config,
}
```

The four backend bools are the ones that really want to collapse, since they all fold into one `BTreeMap<String, bool>` twenty lines down anyway, so threading them separately through three call layers buys nothing. And with `Default` most of the test call sites become `BuildOptionsArgs { cli: &cli, config: &config, ..Default::default() }`, which beats counting falses.

Happy for this to be its own PR if it grows legs, but I'd rather not land the tenth bool and forget about it.

docs/reviews/pr-79.md
```diff
@@ -0,0 +1,61 @@
+# Review handoff: PR #79 — SGLang backend on the shared safetensors substrate
```
deepu105: Can we not commit this one? It names a branch, a PR number and a base SHA, so it's stale the moment this merges. Four of its open items are already in `TODO.md`, which AGENTS.md makes the single index of open work, and now the two can drift apart.

The content is useful, it just belongs in the PR description or as review comments. If part of it is durable, fold that part into `docs/plans/2026-09-09-001-feat-sglang-backend-plan.md`.

deepu105 (commented):
Went through this one properly, mostly the guard and the registration. Really nice work overall. The leak test carrying its own staleness check on the allowlist is a good touch, and `merge_by_path` is a genuine pre-existing bug you found and fixed with a test. The three refactors pull duplication out instead of forking it, which is exactly what I wanted out of the second safetensors engine.

CI is green so I didn't run anything locally. I did go check the pins against SGLang's own source rather than take the description's word for it, and the transcription is exact: all 34 `tool-call-parser` entries match `FunctionCallParser.ToolCallParserEnum` and all 26 `reasoning-parser` entries match `ReasoningParser.DetectorMap` on the v0.5.18 tag, in order, and every declared knob and denylist head is a real `ServerArgs` field. Good.

Two things in the guard I'd like fixed before merge, both cases where the cap silently goes unbounded or unusably small. The rest are follow-ups and nits, all inline.

One thing worth knowing: v0.5.19 shipped on 2026-09-05, four days before this branch. Its breaking changes rename a tool-call parser (`spark` to `spark25`), and our lists are closed `Enum`s, so someone installing SGLang today can get rejected for a name their own server accepts. Not a break since extras still works, but let's re-verify both lists and say in the setup doc which version they came from. I also checked whether 0.5.19's `feat(unified-memory): byte-budget sizing` hands us a byte cap flag and it does not, that's internal pool sizing, so your premise for the token cap still holds.

Last thing, upstream's own help for `--max-total-tokens` says it is "typically used for development and debugging purposes". We're making it the production mechanism on every unified-memory host. I think that's the right call given the alternative, but it's a bet on a flag they don't treat as a supported production knob, so worth a line in `docs/sglang-setup.md` so future us knows it was deliberate.

deepu105:
Thanks @bernardogv really nice work. Left a review.

src/backend/sglang/mod.rs
bernardogv: Done. Both paths now spend their budget through `guard::tokens_for_budget`, which carries the `MIN_POOL_TOKENS` floor, so the unsampled path refuses the same way with `max-total-tokens` named as the override. Added a daemon-driven regression test (`a_pool_under_the_token_floor_is_refused_before_the_host_is_sampled`) with a 16 MB/token config: the 8 GiB default holds 512 tokens, and the launch is refused before spawn.

src/backend/sglang/guard.rs
bernardogv: Fixed in 82907f5. The latent branch is now gated on `rank > 0` and falls through to the head geometry, `num_hidden_layers` is filtered the same way, and any zero product returns `None` (unreadable, refuse). The `max(1)` is gone: the budget-to-tokens step is a separate `tokens_for_budget` that refuses a zero cost outright, so even a caller that hands it 0 cannot get the whole budget as tokens. Two new cases cover `kv_lora_rank: 0` (falls through to heads; `None` when there are none) and zero layers / zero KV heads.

src/backend/sglang/guard.rs
bernardogv: Flipped: `a_zero_per_token_cost_refuses_rather_than_uncapping` asserts `None` from both `max_total_tokens_cap` and `tokens_for_budget`. Verified by reverting the gate and the zero check separately and watching each go red.

src/backend/sglang/mod.rs
bernardogv: Both points confirmed against 0.5.18's `ServerArgs`. For this PR the comment now says what the code does: projected against free rather than the pool, and the gate counts the weights a second time, both over-refusing. The shared fix with the other engine's projection is filed in `TODO.md` and as RV7 in the plan's review section, to land once for both as you suggested.

src/backend/sglang/mod.rs
bernardogv: Agreed, it bounded nothing. `fetch_server_info` now reads the body with `resp.chunk()` and returns `None` as soon as the accumulated length would pass 256 KiB, so memory is bounded and a rejected body is never fed to the parser. Doc comment rewritten to say exactly that. Unit test with a loopback listener: a small body parses, a body past the cap is `None`. I left `orphans::fetch_models_body` as is since you only raised this copy; happy to give it the same treatment in a follow-up.

src/backend/sglang/knobs.rs
bernardogv: Re-verified both maps against the `v0.5.19` tag. One correction to the framing: 0.5.18 has no `spark` entry, so it is not a rename. 0.5.19 adds `dots`, `ling3` and `spark25` to `ToolCallParserEnum` (34 → 37) and `dots`, `ling3` to `DetectorMap` (26 → 28). All five added in map order. `docs/sglang-setup.md` now names the tag and the two source files, and points at the extras tail for a name a newer release accepts before the list is re-transcribed.

src/cli/daemon.rs
bernardogv: Taking you up on the separate PR. Filed in `TODO.md` with your sketch (`BuildOptionsArgs` + `Default`, the four backend bools folded into the `BTreeMap` they become) so it does not get forgotten behind the tenth bool.

docs/reviews/pr-79.md
bernardogv: Removed from the branch. The durable parts moved into `docs/plans/2026-09-09-001-feat-sglang-backend-plan.md`: a "Known limitation" note on identity being minted by whichever safetensors engine claims the path first, and a "Review — PR #79" section with each of these findings and its fix, ticked as landed. The open items were already in `TODO.md`.

deepu105 (approved):
Went through the second round. All eight threads have real code behind them, and I re-checked the parser lists myself against the v0.5.19 tag rather than taking the reply at face value: 37 tool-call and 28 reasoning entries, exact order match, and `choices=["auto"] + map keys` is what `server_args.py` actually does. v0.5.19 is still the newest release, so the lists are current. And you're right that `spark` never existed in 0.5.18, my framing on that was wrong.

The zero-cost geometry fix and the unsampled floor are both what I was after, and dropping the local duplicate-id test is fine now that #78's validator walks every backend's knobs.

One thing from my review didnt make it: the note that upstream calls `--max-total-tokens` a development and debugging flag while we make it the production mechanism on every unified-memory host. Still worded that way in 0.5.19. Not worth another round trip, I'll add it myself.

I also turned up a few new things on this pass. Fixing them directly on main rather than sending this back:

- The two new refusal tests race the host metrics sampler. `resolve_knobs` bails out with no cap once the host is sampled and not unified, and the first sample lands about a second after daemon start, so on a CPU-only runner both `expect_err` tests flip to successful launches and fail.
- The floor test cant tell the two paths apart. Both refusal strings carry `max-total-tokens` and `2048 tokens`, so it passes either way.
- `merge_by_path` sorts projectors by launch priority, but the auto route for a safetensors path comes from `synthetic_identity_for_path`, which walks registration order. Those agree today by luck, not by construction.
- `knob_f64`, `user_set` and `adoption_matches` are verbatim copies of the vLLM ones. Fourth thing that wants sharing, alongside RV7.

I'm also going to land the `BuildOptionsArgs` cleanup (RV8) on main now rather than leave it queued, since the tenth bool is in the tree as of this merge.

Merging. Thanks for the spark2 load test record, and for keeping each refactor as its own commit, that made this much easier to review.

PR #81, opened 2026-09-10
fix(unit5): the vLLM UMA guard launches beside a tenant and reserves for the engine's own overhead

### LINKED ISSUE #80: vLLM UMA guard on GB10: the byte cap can't launch, and the 8 GiB reserve is spent on engine overhead
Following up on your "appreciate any more testing on NVidia hardware as I'm on AMD stack" from #60 — I put the vLLM unified-memory guard through a DGX Spark (GB10). Since #60 this box reports `unified: true`, so it takes the same path as a Strix Halo, but as far as I can tell nobody had actually run it on NVIDIA coherent UMA.

Two gaps, both measured. There's also one piece of good news, so let me lead with that.

**`effective_free_bytes` is correct here.** Five back-to-back samples against `/proc/meminfo` agreed within 35 MB, classified `IntegratedUma` at fraction 1.0. `sysinfo::used_memory()` is `MemTotal - MemAvailable` on Linux, so the subtraction lands on `MemAvailable`, which is the only trustworthy figure on this platform — raw `cudaMemGetInfo` reported 2.77 GiB free against `MemAvailable` 52 GiB at the same instant. No change needed.

**Setup.** GB10 / DGX Spark, 121.69 GiB unified LPDDR5X, swap off, driver 580.173.02, `vllm/vllm-openai:v0.28.0` by digest. llamastash at `feat/sglang-backend` @ `e574b3a` — the guard arithmetic there is identical to `main`, only the constant's home differs. Isolated daemon, `/proc/meminfo` sampled at 1 Hz for every run.

---

### 1. The byte cap can't launch vLLM 0.28 unless the box is ~92% free

With one other tenant resident (52 GiB available), llamastash emitted `--kv-cache-memory-bytes 8589934592` exactly as expected, and vLLM refused at `init_device`, before touching weights:

```
ValueError: Free memory on device cuda:0 (49.84/121.69 GiB) on startup is less than desired
GPU memory utilization (0.92, 111.95 GiB). Decrease GPU memory utilization or reduce GPU
memory used by other processes.
```

`request_memory()` in `vllm/v1/worker/utils.py:444` compares free against `total × gpu_memory_utilization` unconditionally:

```python
requested_memory = math.ceil(init_snapshot.total_memory * cache_config.gpu_memory_utilization)
if init_snapshot.free_memory < requested_memory:
    raise ValueError(...)
```

`--kv-cache-memory-bytes` isn't consulted until `determine_available_memory()`, much later — where vLLM does honour it, skips profiling, and logs that the cap "does not respect the gpu_memory_utilization config".

So the guard's argv is launchable only on a UMA host that is ≥92% free at the moment the CUDA context exists. On an idle box it passed by 1.81 GiB. On an empty Strix Halo that's presumably always true, which would explain why this hasn't surfaced — but with any co-tenant it never is.

**Fix shape:** emit `--gpu-memory-utilization` alongside the cap, sized so `total × util <= free` at launch — e.g. `(weights + cap + reserve) / total`. With the cap set, vLLM ignores `util` for everything else, so it only has to clear the startup check.

### 2. The 8 GiB host reserve is consumed by engine overhead

`UNIFIED_HOST_RESERVE_BYTES` is documented as what's "left free for the OS and everything else after weights + cache". Measured on this box, the byte-cap path costs well over half of it before the OS sees anything:

| model | launch cost | weights + cap | overhead |
|---|---|---|---|
| Qwen2.5-0.5B | 14.33 GiB | 0.93 + 8.00 | **5.40 GiB** |
| Nemotron-3.5-Lightning-30B-A3B (NVFP4) | 34.81 GiB | 20.08 + 8.00 | **6.73 GiB** |

It isn't a constant — the difference is CUDA-graph memory, 1.0 → 3.0 GiB. So in the regime where the reserve actually binds:

```
MemAvailable_at_ready  =  8 GiB - overhead  ≈  1.3 GiB
```

on a 121.69 GiB machine. I measured that rather than projecting it. Driving free memory down so the cap bound at 4.29 GiB, with a sacrificial process as the designated OOM victim:

| moment | MemAvailable |
|---|---|
| after weights loaded, steady | 9.2 GiB |
| vLLM reserves the 4.29 GiB KV pool | **6.40 GiB (1 Hz sample)** — `earlyoom` fired: `at or below SIGTERM limits: mem 5.00%` |
| CUDA-graph capture then took a further 3.01 GiB | would have landed ~0.9 GiB |

The single KV reservation crossed the 5% line, before graph capture had even started. This box runs `earlyoom`, so it killed the sacrificial process; a stock GB10 without it hits a documented hard-hang-and-reboot failure, because on unified memory an over-allocation doesn't surface as a clean CUDA OOM.

### Why these are one fix

Gap 1 is currently masking gap 2. The reserve only binds when `free < weights + 16 GiB`, and vLLM's startup check needs `free ≳ 114.6 GiB` — so on a 121.69 GiB box only a >98 GiB model could reach the binding regime, and the guard refuses those outright. **Fixing gap 1 alone would expose gap 2 to every user.**

### The question

I didn't want to arrive with a patch, because this is your safety constant and any change to it lands on AMD hardware I can't test. Two shapes seem plausible:

- **(a)** `reserve = OS floor + the backend overhead band the admission gate already computes`, so it tracks reality rather than a magic number. Caveat: the measured overhead isn't constant (5.4 → 6.7 GiB with graph memory), so the existing band may not model it either.
- **(b)** a fraction of `MemTotal` — 12–18% — so small and large APUs both land sensibly. On this box, 18% (~22 GiB) keeps ~15 GiB free at ready; 12% (~14 GiB) just clears a 5% OOM-killer line.

I'd lean (a) with (b) as the fallback, but you know the AMD side and I don't.

One thing I want to be careful about: **I have not measured any of this on a Strix Halo**, so I'm not claiming your box is broken. Gap 2 is structural rather than NVIDIA-specific — engine overhead comes out of the reserve regardless of vendor — but whether it bites on AMD depends on numbers I don't have. Worth a check on your side.

Happy to do the implementation and re-run the full matrix on GB10 once you've picked a direction. Raw traces and per-run logs available if any of it would be useful.

bernardogv:
Fixes #80. Both gaps in one PR because gap 1 was masking gap 2: with the cap only launchable on a ~92%-free host, the reserve could never bind on a 121 GiB box unless the model was >98 GiB, which the guard refuses anyway.

## Gap 1 — the byte cap could not launch beside a tenant

vLLM 0.28's `request_memory()` (`vllm/v1/worker/utils.py:444`, v0.28.0 tag) compares `init_snapshot.free_memory` against `total × gpu_memory_utilization` unconditionally, before weights; `--kv-cache-memory-bytes` is only honoured later in `determine_available_memory()`. So the guard's argv started only on a host that was ~92% free at the moment the CUDA context existed.

`startup_utilization` now rides alongside the cap:

```
demand       = weights + cap + VLLM_ENGINE_OVERHEAD_BYTES
visible_free = MemAvailable − VLLM_CONTEXT_MARGIN_BYTES        # 3 GiB
utilization  = floor(min(demand, visible_free) / total, 2 dp), clamped to [0.01, 0.99]
```

The context margin matters: vLLM's own free reading ran 2.6–2.7 GiB under MemAvailable on GB10 (113.69 vs 116.43; 29.83 vs 32.40) because the container and CUDA context are already spent when it samples. Sized against MemAvailable alone, the tight case rounds to 0.25 and asks for 30.4 GiB against 29.83 visible, and vLLM still refuses. The unit test pins both measured runs against the free figure vLLM actually reported. Once the cap is set the fraction governs nothing else.

## Gap 2 — the 8 GiB reserve was spent on engine overhead

Measured launch cost minus weights minus cap on GB10: **5.40 GiB** (Qwen2.5-0.5B) and **6.73 GiB** (Nemotron-3.5-Lightning-30B-A3B NVFP4), the difference being CUDA-graph memory (1.0 → 3.0 GiB). In the binding case that leaves `8 − 6.7 ≈ 1.3 GiB` MemAvailable at ready; measured live, `earlyoom` fired the instant vLLM reserved a 4.29 GiB pool, before graph capture.

Per your "(a) with (b) as fallback":

```
reserve = max( OS floor 2 GiB + engine overhead 7 GiB + overhead_band_bytes(gpu_backend),
               15% × MemTotal )
```

**One honest caveat on (a) as written.** The band the admission gate already computes is `overhead_band_bytes`: 512 MiB native, 1 GiB unknown. That is an order of magnitude under the measured 5.4–6.7 GiB, so "reuse the existing band" cannot model this on its own. The 7 GiB engine term is the measured ceiling rounded up, from two models on one box with graph memory spanning 1.0–3.0 GiB. It is not a constant of nature. It is a named constant with its provenance in the doc comment, and the escape hatch is an explicit `kv_cache_memory_bytes`, which still opts out of the guard entirely. If you would rather it were a config key, say so and I will add one.

Worked examples, so you can sanity-check your own boxes:

| pool | fixed terms | 15% of pool | reserve | was |
|---|---|---|---|---|
| 32 GiB APU | 9.5 GiB | 4.8 GiB | **9.5 GiB** (30%) | 8 GiB |
| 64 GiB APU | 9.5 GiB | 9.6 GiB | **9.6 GiB** | 8 GiB |
| 121.69 GiB (GB10 / Strix Halo 128) | 9.5 GiB | 18.25 GiB | **18.25 GiB** | 8 GiB |

On the small APU the change is +1.5 GiB and the constant dominates; on the large ones the fraction dominates and lands ~11.5 GiB MemAvailable at ready in the binding case (18.25 − 6.7). 12% would land ~7.9 GiB, just over a 5% OOM-killer line; 18% ~15 GiB. 15% is the middle of your range and is one number to tune.

## What I would like back from the AMD side

- Whether the 7 GiB engine overhead and the 3 GiB context margin hold on Strix Halo. The two numbers to compare: `launch cost − weights − cap` at ready, and `MemAvailable − vLLM's logged init free`.
- Whether a capped launch beside a tenant now starts (the run A shape: another process holding ~half the box).

## Not in this PR

- The unsampled path (first launch after daemon start) still sends the default 8 GiB cap with no utilization: there is no pool total to size one against. Narrow window; in `TODO.md`.
- #79 hoists the reserve constant into `launch::admission` and SGLang's token-cap guard shares it. This branch is off `main`; if #79 merges first the same change moves into `admission.rs` and covers SGLang for free (the SGLang UAT measured ~8 GiB of non-pool footprint on the same box, so gap 2 applies there verbatim). If this merges first, I will rebase #79 onto it.

## Verification

`make test` and `make lint` green. Each of the three new pieces was broken in turn (margin dropped, fraction floor dropped, engine overhead left out of the utilization) and its test went red. Not yet re-run on GB10 against the merged binary; that is the second `TODO.md` item and I will do it once this lands.

Evidence: [issue #80](https://github.com/llamastash/llamastash/issues/80); raw traces in the casebook record cited there.


deepu105:
Rebased this onto main now that #79 is in, reviewed it, and pushed the fixes as a second commit rather than sending it back.

On the rebase, I took the option you offered: the reserve, the OS floor and the 15% fraction now live in `launch::admission`, with the engine overhead as a **parameter** rather than a constant in there. That keeps the no-leak rule intact (my first attempt named vLLM in the shared doc comments and the leak guard in `vllm_backend_test.rs` caught it, which was satisfying to watch). `unified_kv_cache_budget` takes the reserve now.

One thing I did not do: put SGLang on the new reserve. You said gap 2 applies verbatim, and I believe it, but the number I'd need isn't in the record. The spark2 UAT's "22 GiB of 121 GiB peak" is a whole-host figure, not `launch cost - weights - pool` the way your 5.4/6.7 GiB are, so adopting 7 GiB for SGLang would be moving a live number on a guess. It keeps passing the flat 8 GiB constant, so its behaviour is byte-for-byte what it was, and the measurement plus the one constant is in `TODO.md`. The function already takes the parameter, so it's a two-line change once someone measures it.

## The one real fix

Copilot caught this and it's correct: **an explicit `kv_cache_memory_bytes` never got a companion utilization.** It hit the early return in `resolve_knobs`, so the user's own cap still met the untouched 0.92 startup check and got refused beside a tenant. That's the failure this PR exists to fix, and the explicit cap is the escape hatch both `docs/vllm-setup.md` and `docs/troubleshooting.md` point users at for exactly this host, so it was broken where it mattered most.

Now only a user-set *fraction* opts out entirely. A user-set cap keeps its value verbatim and still gets a utilization derived from it.

Testing that needed a sampled unified host, which a test daemon never is (the sampler's first snapshot lands a second in, and a runner with no GPU samples as `cpu_only`). `MethodContext::host_metrics` is a public field, so the new test hands `resolve_knobs` a hand-built snapshot directly. That's the only way to reach the branch, and it's worth knowing the integration suite structurally cannot.

## Smaller things

- The setup guide claimed the companion fraction unconditionally. The unsampled branch has no pool total to size one against, so both docs now name that exception. Copilot's third comment, also fair.
- The utilization write ignored `set_by_name_for`'s result. It can't fail today, but a silent no-op there puts the cap back in front of the 0.92 check, which is the whole bug, so it logs now.
- The "tight host" numbers in `startup_utilization_clears_the_init_check_against_what_vllm_sees` are labelled as a branch exercise rather than a reachable state. Worth checking my arithmetic: with the engine-aware reserve, 15% of 121.69 GiB is 18.25 GiB, so 32.4 free minus 20.1 weights leaves nothing and `unified_kv_cache_budget` refuses that host outright. Gap 2's fix makes gap 1's tight case unreachable, the mirror of the masking you described. The co-tenant case (52 GiB free) still goes through the fraction path, which is the one that matters. `a_host_the_reserve_refuses_never_reaches_the_utilization` pins it.

## Left for later

Copilot's suppressed comment is a real one and I'd rather not rush it: **the admission ledger prices a launch without the engine overhead.** `launch_service.rs` uses weights + `projected_cache_bytes` + `overhead_band_bytes` (512 MiB), so it never sees the 5.4-6.7 GiB you measured. One launch is fine, the reserve already carved that out of the cap. Two concurrent launches read the same host sample and each reserve about 6.5 GiB less than they take, which is precisely what the ledger is for.

The trait doc for `projected_cache_bytes` says "bytes this launch will hold beyond its weights", so the overhead belongs there by contract. What stops it being a one-liner is that the hook gets no host snapshot, and adding 7 GiB unconditionally would over-refuse on a discrete card where nobody has measured it. So it wants either the snapshot threaded in or a separate `engine_overhead_bytes` the gate adds only on a unified host. In `TODO.md` with that reasoning.

`make test` and `make lint` green, 47 test binaries. Merging once CI is. Two notes for when you re-run the GB10 matrix: the explicit-cap path is new behaviour worth a run of its own, and the 7 GiB and 3 GiB figures still want a second box.

PR #82, opened 2026-09-11
fix(admission): use AMD GTT budget for UMA GPUs inside LXC

Ramon-Balaguer:
## Summary

Fix memory admission for AMD UMA GPUs exposed through Linux LXC containers.

On an AMD Ryzen AI Max+ / Radeon 8060S system, llama.cpp/ROCm can see the GPU's full GTT pool (96 GiB in the reported setup), while `/proc/meminfo` inside the LXC reports only the container's configured RAM (8 GiB). The existing UMA admission policy uses `min(ram_free, gtt_free)`, so LlamaStash incorrectly refuses models that fit in the GPU pool.

This PR narrows the exception to:

- Linux LXC
- AMD backend
- unified/UMA GPU
- GTT metrics available

For that combination, admission budgets the AMD GTT pool directly. All other UMA paths retain the existing conservative `min(ram_free, gtt_free)` behavior.

## Why this is safe

The change does not alter llama.cpp placement, `--fit`, headroom calculation, or the reservation ledger. It only changes the admission budget used for the GPU pool in the specific LXC+AMD+UMA case.

The container's RAM limit still applies to CPU-side allocations; the GTT pool is used as the GPU allocation budget. The observed environment validates this distinction: llama.cpp reports ~96 GiB available on the Radeon 8060S, and forcing a 27B Q8 model through the existing gate results in approximately 47 GiB of actual AMD VRAM/GTT usage.

## Implementation

- Add local Linux LXC detection using `/run/systemd/container`, PID 1 environment, and cgroup fallbacks.
- Add a small policy helper so the container-specific behavior is explicit and unit-testable.
- Keep the existing RAM/GTT minimum for bare-metal AMD/Intel UMA, Apple Silicon, other containers, and missing GTT data.

## Tests

Added coverage for:

- AMD UMA in LXC uses GTT instead of the container RAM budget.
- Bare-metal AMD UMA retains the RAM/GTT minimum.
- Non-AMD LXC UMA does not use the GTT-only exception.
- AMD UMA without GTT data still falls back to RAM.

Expected validation in CI: `cargo fmt --check`, `cargo clippy -- -D warnings`, and `cargo test --features test-fixtures`.

### DIFF
--- a/src/launch/admission.rs
+++ b/src/launch/admission.rs
@@ -264,8 +264,6 @@ pub fn effective_free_bytes(snap: &HostMetricsSnapshot) -> u64 {
 
 fn effective_free_bytes_for(snap: &HostMetricsSnapshot, lxc: bool) -> u64 {
   let ram_free = snap.ram_total_bytes.saturating_sub(snap.ram_used_bytes);
-  // Apple is unified by construction (the `|| apple_metal` just guards
-  // it); the host-pane VRAM gauge keys off the same `unified` flag.
   let unified = snap.unified || snap.gpu_backend == HostMetricsSnapshot::BACKEND_APPLE_METAL;
   if unified {
     let pool_free = match snap.uma_shared_total_bytes {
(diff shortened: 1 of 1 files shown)


deepu105:
Thank you @Ramon-Balaguer let me know when the PR is ready

Ramon-Balaguer:
@deepu105 I want create an other PR

PR #83, opened 2026-09-12
fix(admission): use AMD GTT budget for UMA GPUs inside LXC

Ramon-Balaguer:
## Summary

Fix memory admission for AMD UMA GPUs exposed through Linux LXC containers.

On an AMD Ryzen AI Max+ / Radeon 8060S system, llama.cpp/ROCm can see the GPU's full GTT pool (96 GiB in the reported setup), while `/proc/meminfo` inside the LXC reports only the container's configured RAM (8 GiB). The existing UMA admission policy uses `min(ram_free, gtt_free)`, so LlamaStash incorrectly refuses models that fit in the GPU pool.

This PR narrows the exception to:

- Linux LXC
- AMD backend
- unified/UMA GPU
- GTT metrics available

For that combination, admission budgets the AMD GTT pool directly. All other UMA paths retain the existing conservative `min(ram_free, gtt_free)` behavior.

## Why this is safe

The change does not alter llama.cpp placement, `--fit`, headroom calculation, or the reservation ledger. It only changes the admission budget used for the GPU pool in the specific LXC+AMD+UMA case.

The container's RAM limit still applies to CPU-side allocations; the GTT pool is used as the GPU allocation budget. The observed environment validates this distinction: llama.cpp reports ~96 GiB available on the Radeon 8060S, and forcing a 27B Q8 model through the existing gate results in approximately 47 GiB of actual AMD VRAM/GTT usage.

## Implementation

- Add local Linux LXC detection using `/run/systemd/container`, PID 1 environment, and cgroup fallbacks.
- Add a small policy helper so the container-specific behavior is explicit and unit-testable.
- Keep the existing RAM/GTT minimum for bare-metal AMD/Intel UMA, Apple Silicon, other containers, and missing GTT data.

## Tests

Added coverage for:

- AMD UMA in LXC uses GTT instead of the container RAM budget.
- Bare-metal AMD UMA retains the RAM/GTT minimum.
- Non-AMD LXC UMA does not use the GTT-only exception.
- AMD UMA without GTT data still falls back to RAM.

Expected validation in CI: `cargo fmt --check`, `cargo clippy -- -D warnings`, and `cargo test --features test-fixtures`.

### DIFF
--- a/src/launch/admission.rs
+++ b/src/launch/admission.rs
@@ -191,33 +191,74 @@ pub fn dir_weight_bytes(dir: &std::path::Path) -> u64 {
     .fold(0u64, u64::saturating_add)
 }
 
-/// Post-headroom free bytes across the budget pool(s). Discrete hosts
-/// sum post-headroom VRAM free + post-headroom system-RAM free.
-///
-/// UMA hosts budget the **GPU pool**, not all of system RAM. On an
-/// AMD/Intel integrated APU the GPU can only allocate within the amdgpu
-/// GTT cap (carve-out + GTT), which on a default-config box is roughly
-/// half of system RAM. llama.cpp's own free reading conflates the two
-/// and hard-OOMs (it sees system-RAM free, allocates past the GTT cap,
-/// and `hipMalloc` fails); sysfs GTT is the budget authority. So when
-/// the snapshot carries the GTT pool (`uma_shared_*`, from the sysfs
-/// probe) we budget `min(ram_free, gtt_free)` — the GTT cap bounds the
-/// GPU allocation, and `ram_free` still guards the rare case where
-/// system RAM is the tighter constraint. Apple Silicon has no GTT carve
-/// (it leaves `uma_shared_*` unset), so it falls back to `ram_free` with
-/// its 0.75 headroom.
-pub fn effective_free_bytes(snap: &HostMetricsSnapshot) -> u64 {
+/// Detect an AMD/Intel unified-memory Linux container running inside an LXC.
+/// This is the specific case where the container sees an artificially low
+/// host RAM floor but the GPU stack still operates against the real GTT pool.
+fn linux_lxc_container() -> bool {
+  if !cfg!(target_os = "linux") {
+    return false;
+  }
+
+  if let Ok(contents) = std::fs::read_to_string("/run/systemd/container") {
+    let normalized = contents.trim().to_ascii_lowercase();
+    if !normalized.is_empty() && (normalized == "lxc" || normalized.contains("lxc")) {
+      return true;
+    }
+  }
+
+  if let Ok(env) = std::fs::read_to_string("/proc/1/environ") {
+    let lower = env.to_ascii_lowercase();
+    if lower.contains("lxc")
+      || lower.contains("container=lxc")
+      || lower.contains("container=lxc-libvirt")
+    {
+      return true;
+    }
+  }
+
+  if let Ok(cgroup) = std::fs::read_to_string("/proc/1/cgroup") {
+    let lower = cgroup.to_ascii_lowercase();
+    if lower.contains("lxc") || lower.contains("/lxc/") || lower.contains("lxc.monitor") {
+      return true;
+    }
+  }
+
+  false
+}
+
+/// AMD UMA in Linux LXC can expose a real GTT pool even when the container RAM
+/// limit is much smaller. In that specific case the GPU pool is the budget
+/// authority; everywhere else keep the conservative `min(ram_free, gtt_free)`
+/// policy.
+fn amd_uma_lxc_pool_free(snap: &HostMetricsSnapshot, in_lxc: bool) -> Option<u64> {
+  if !in_lxc
+    || snap.gpu_backend != HostMetricsSnapshot::BACKEND_AMD
+    || !snap.unified
+    || snap.uma_shared_total_bytes.is_none()
+  {
+    return None;
+  }
+
+  let gtt_total = snap.uma_shared_total_bytes?;
+  Some(gtt_total.saturating_sub(snap.uma_shared_used_bytes.unwrap_or(0)))
+}
+
+fn effective_free_bytes_for_policy(snap: &HostMetricsSnapshot, in_lxc: bool) -> u64 {
   let ram_free = snap.ram_total_bytes.saturating_sub(snap.ram_used_bytes);
   // Apple is unified by construction (the `|| apple_metal` just guards
   // it); the host-pane VRAM gauge keys off the same `unified` flag.
   let unified = snap.unified || snap.gpu_backend == HostMetricsSnapshot::BACKEND_APPLE_METAL;
   if unified {
-    let pool_free = match snap.uma_shared_total_bytes {
-      Some(gtt_total) => {
-        let gtt_free = gtt_total.saturating_sub(snap.uma_shared_used_bytes.unwrap_or(0));
-        ram_free.min(gtt_free)
+    let pool_free = if let Some(gtt_free) = amd_uma_lxc_pool_free(snap, in_lxc) {
+      gtt_free
+    } else {
+      match snap.uma_shared_total_bytes {
+        Some(gtt_total) => {
+          let gtt_free = gtt_total.saturating_sub(snap.uma_shared_used_bytes.unwrap_or(0));
+          ram_free.min(gtt_free)
+        }
+        None => ram_free,
       }
-      None => ram_free,
     };
     admissible_bytes(pool_free, pool_kind(snap))
   } else if let (Some(total), Some(used)) = (snap.gpu_mem_total_bytes, snap.gpu_mem_used_bytes) {
@@ -227,6 +268,12 @@ pub fn effective_free_bytes(snap: &HostMetricsSnapshot) -> u64 {
   } else {
     admissible_bytes(ram_free, PoolKind::SystemRam)
   }
+}
+
+/// Post-headroom free bytes across the budget pool(s). Discrete hosts
+/// sum post-headroom VRAM free + post-headroom system-RAM free.
+pub fn effective_free_bytes(snap: &HostMetricsSnapshot) -> u64 {
+  effective_free_bytes_for_policy(snap, linux_lxc_container())
 }
 
 /// Demand floor for a launch: model weights + KV cache at the effective
@@ -433,6 +480,16 @@ mod tests {
   }
 
   #[test]
+  fn amd_lxc_uma_uses_gtt_budget_instead_of_container_ram() {
+    let mut s = snap(HostMetricsSnapshot::BACKEND_AMD, true, 160 * GIB, 80 * GIB);
+    s.uma_shared_total_bytes = Some(80 * GIB);
+    s.uma_shared_used_bytes = Some(60 * GIB);
+    assert_eq!(amd_uma_lxc_pool_free(&s, true), Some(20 * GIB));
+    assert_eq!(effective_free_bytes_for_policy(&s, true), 20 * GIB);
+    assert_eq!(effective_free_bytes_for_policy(&s, false), 20 * GIB);
+  }
+
+  #[test]
   fn uma_budget_uses_gtt_pool_not_system_ram() {
     // Default-config UMA box: the amdgpu GTT cap is ~half of system RAM.
     // A resident model leaves plenty of system RAM free but little GTT.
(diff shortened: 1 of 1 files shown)


Ramon-Balaguer:
@deepu105 Hi! You can check it now.

Ramon-Balaguer:
# Summary
This push includes commit b383a78:

- Fixes double-counting of memory on AMD UMA systems when vram_total already includes the full GTT pool.
- Preserves the full GTT pool for the VRAM gauge in Linux LXC containers.
- Centralizes LXC container detection in process.rs.
- Updates model admission logic to use the available memory correctly.
- Adds coverage for unified-memory APUs, including the 96 GiB GTT case.

@deepu105

deepu105:
Thanks @Ramon-Balaguer I'll reveiw and merge soon.

src/gpu/mod.rs
```diff
@@ -79,13 +79,20 @@ pub fn classify_amd_memory(
...
-      ClassSource::CarveSignature,
-    )
+    // When vram_total is already large (e.g. equal to GTT on APUs like Strix Halo
+    // or when the driver reports the full aperture in vram_total), summing them
+    // would double-count the same physical unified memory pool.
+    let total = if vram_total >= gt {
+      vram_total
+    } else {
+      vram_total.saturating_add(gt)
+    };
```
deepu105, line 89: `mem_info_vram_total` and `mem_info_gtt_total` are separate amdgpu pools (the BIOS carve and the system-RAM-backed GTT), so `vram_total >= gtt_total` doesn't mean the same memory is counted twice. `docs/usage.md` documents `gpu_pool_total_bytes` as carve-out + GTT for this reason.

Strix Halo with the stock 4 GiB carve:

```
mem_info_vram_total   4294967296
mem_info_gtt_total  130452963328
```

That host takes the `else` branch and sums, which is correct. On a host with a large BIOS carve (`vram_total >= gtt_total`), this branch drops the GTT pool from the total.

If the LXC host reports a GTT total larger than the RAM behind it, clamping GTT against that RAM would fix it without skipping the sum. This change also applies to bare-metal hosts and isn't in the PR description or commit messages, so it should move out of this PR unless the sysfs values show a real double count.

src/gpu/mod.rs
```diff
@@ -79,13 +79,20 @@ pub fn classify_amd_memory(
...
+    let used = if vram_total >= gt && vram_used >= gu {
+      vram_used
+    } else {
+      vram_used.saturating_add(gu)
+    };
```
deepu105, line 94: `total` and `used` pick their branch with different conditions, so `used` can be larger than `total`. With `vram_total >= gt` but `vram_used < gu`:

```rust
classify_amd_memory(96 * GIB, 50 * GIB, Some(96 * GIB), Some(90 * GIB), true)
// total = 96 GiB, used = 140 GiB
```

Free then saturates to 0 and the VRAM gauge sits at 100%. If this change stays, both need the same condition.

src/launch/admission.rs
```diff
@@ -572,6 +584,16 @@ mod tests {
...
   }
 
+  #[test]
+  fn amd_lxc_uma_uses_gtt_budget_instead_of_container_ram() {
+    let mut s = snap(HostMetricsSnapshot::BACKEND_AMD, true, 160 * GIB, 80 * GIB);
+    s.uma_shared_total_bytes = Some(80 * GIB);
+    s.uma_shared_used_bytes = Some(60 * GIB);
+    assert_eq!(amd_uma_lxc_pool_free(&s, true), Some(20 * GIB));
+    assert_eq!(effective_free_bytes_for_policy(&s, true), 20 * GIB);
+    assert_eq!(effective_free_bytes_for_policy(&s, false), 20 * GIB);
```
deepu105, line 594: This test passes without the LXC exception. With `ram_free = 80 GiB` and `gtt_free = 20 GiB`, the default `min(ram_free, gtt_free)` already gives 20 GiB, so the `true` and `false` asserts check the same value. With `amd_uma_lxc_pool_free` changed to always return `None`, only the helper assert on line 592 fails.

The container case needs container RAM below GTT free. This one passes on the branch and fails without the exception:

```rust
let mut s = snap(HostMetricsSnapshot::BACKEND_AMD, true, 8 * GIB, 2 * GIB);
s.uma_shared_total_bytes = Some(96 * GIB);
s.uma_shared_used_bytes = Some(0);
assert_eq!(effective_free_bytes_for_policy(&s, true), 96 * GIB);
assert_eq!(effective_free_bytes_for_policy(&s, false), 6 * GIB);
```

src/launch/admission.rs
```diff
@@ -270,33 +270,39 @@ pub fn dir_weight_bytes(dir: &std::path::Path) -> u64 {
...
   let ram_free = snap.ram_total_bytes.saturating_sub(snap.ram_used_bytes);
   // Apple is unified by construction (the `|| apple_metal` just guards
   // it); the host-pane VRAM gauge keys off the same `unified` flag.
   let unified = snap.unified || snap.gpu_backend == HostMetricsSnapshot::BACKEND_APPLE_METAL;
   if unified {
-    let pool_free = match snap.uma_shared_total_bytes {
-      Some(gtt_total) => {
-        let gtt_free = gtt_total.saturating_sub(snap.uma_shared_used_bytes.unwrap_or(0));
-        ram_free.min(gtt_free)
+    let pool_free = if let Some(gtt_free) = amd_uma_lxc_pool_free(snap, in_lxc) {
```
deepu105, line 296: This conflicts with main in this file, and resolving the text won't be enough. Main now has `engine_pool_total_bytes`, which returns `ram_total_bytes` on unified hosts. vLLM `gpu_memory_utilization` and SGLang `mem_fraction_static` are priced as a share of that pool (`pool_fraction_beyond_weights`).

In an 8 GiB container with this change:

- `effective_free_bytes`: ~96 GiB (GTT free)
- `engine_pool_total_bytes`: 8 GiB (container `MemTotal`)

A `0.9` fraction is priced at 7.2 GiB or less and admitted against 96 GiB free. If torch on ROCm reports the ~96 GiB pool as its device total (not measured on an ROCm APU yet), the engine takes ~86 GiB. After the rebase both functions should use the same pool in the container case.

src/launch/admission.rs
```diff
@@ -270,33 +270,39 @@ pub fn dir_weight_bytes(dir: &std::path::Path) -> u64 {
...
-/// GPU allocation, and `ram_free` still guards the rare case where
-/// system RAM is the tighter constraint. Apple Silicon has no GTT carve
-/// (it leaves `uma_shared_*` unset), so it falls back to `ram_free` with
-/// its 0.75 headroom.
-pub fn effective_free_bytes(snap: &HostMetricsSnapshot) -> u64 {
+/// AMD UMA in Linux LXC can expose a real GTT pool even when the container RAM
+/// limit is much smaller. In that specific case the GPU pool is the budget
+/// authority; everywhere else keep the conservative `min(ram_free, gtt_free)`
+/// policy.
+fn amd_uma_lxc_pool_free(snap: &HostMetricsSnapshot, in_lxc: bool) -> Option<u64> {
```
deepu105, line 277: `effective_free_bytes` is the budget for every launch on the host, not only fully offloaded ones. In the container case this drops `ram_free` completely, so a CPU-only or partly offloaded launch in an 8 GiB container is checked against ~96 GiB. The PR description notes the container RAM limit still applies to CPU-side allocations, so those launches can pass the gate and then hit the 8 GiB limit.

src/launch/admission.rs
```diff
@@ -308,6 +314,12 @@ pub fn effective_free_bytes(snap: &HostMetricsSnapshot) -> u64 {
   }
 }
 
+/// Post-headroom free bytes across the budget pool(s). Discrete hosts
+/// sum post-headroom VRAM free + post-headroom system-RAM free.
```
deepu105, line 318: The UMA explanation that was on this doc comment got removed (why sysfs GTT is the budget, the `hipMalloc` OOM, Apple having no GTT), and the new text only covers discrete hosts. Please move it back here or onto `effective_free_bytes_for_policy`, and add the container exception to it.

src/tui/host_stats_pane.rs
```diff
@@ -213,10 +213,35 @@ fn gpu_summary_row<'a>(
...
 /// the full pool when RAM is free, dropping toward free-RAM headroom under
 /// pressure. Clamped `>= used` so the bar never exceeds 100%.
+/// In an AMD Linux LXC container where container RAM limit is artificial,
+/// the real GPU pool is preserved without clamping against container RAM.
 fn vram_denominator(host: &HostMetricsSnapshot, used: u64, total: u64) -> u64 {
+  vram_denominator_for_policy(
+    host,
+    used,
+    total,
+    crate::util::process::linux_lxc_container(),
```
deepu105, line 223: `vram_denominator` runs from `vram_row` on every draw of the host pane, so this reads `/run/systemd/container`, `/proc/1/environ` and `/proc/1/cgroup` each frame. Being in a container doesn't change at runtime, so `linux_lxc_container()` can cache its result in a `OnceLock`. Admission calls it on each launch too.

src/util/process.rs
```diff
@@ -144,6 +144,65 @@ fn terminate_child_tree(child: &mut Child) {
...
+    {
+      return true;
+    }
+  }
+
+  if let Some(cgroup) = proc_cgroup {
+    let lower = cgroup.to_ascii_lowercase();
+    if lower.contains("lxc") || lower.contains("/lxc/") || lower.contains("lxc.monitor") {
+      return true;
+    }
```
deepu105, line 200: Each block reduces to `contains("lxc")`, which already covers the other checks (`normalized == "lxc"`, `container=lxc`, `container=lxc-libvirt`, `/lxc/`, `lxc.monitor`). So any `lxc` substring matches, for example this returns `true`:

```rust
is_lxc_container_signatures(None, Some("PATH=/usr/bin:/opt/lxc-tools/bin\0HOME=/root"), None)
```

`/proc/1/environ` is only readable by root, so this needs the daemon running as root, but the checks should be exact: a whole `container=lxc` or `container=lxc-libvirt` entry in the `\0`-separated environ, and the trimmed `/run/systemd/container` value equal to `lxc` or `lxc-libvirt`.

src/util/process.rs
```diff
@@ -144,6 +144,65 @@ fn terminate_child_tree(child: &mut Child) {
   let _ = child.kill();
 }
 
+/// Detect whether a root filesystem contains Linux LXC container signatures.
+///
+/// Inspects (relative to `root`):
+/// 1. `run/systemd/container`
+/// 2. `proc/1/environ`
+/// 3. `proc/1/cgroup`
+pub fn linux_lxc_container_at(root: &std::path::Path) -> bool {
```
deepu105, line 153: `linux_lxc_container_at` and `is_lxc_container_signatures` are only used inside the crate, `pub(crate)` is enough for both.

CHANGELOG.md
```diff
@@ -4,12 +4,8 @@ All notable changes to LlamaStash will be documented in this file. The format fo
 
 ## [Unreleased]
 
-### Added
-
+- **TUI Host pane VRAM gauge in Linux LXC containers on AMD UMA.** When running inside a Linux LXC container where container RAM is artificially limited, the VRAM denominator preserves the host's full GTT pool rather than clamping against the container's RAM limit.
```
deepu105, line 7: This removes the `### Added` and `### Fixed` headers under `[Unreleased]`, please keep them. The entry also only covers the VRAM gauge. The admission change is user-visible too and belongs under `### Fixed`, for example: launches inside an LXC container on AMD UMA are checked against the GTT pool instead of the container RAM limit.

deepu105 (commented):
**AI review.** Thanks for picking this up @Ramon-Balaguer, running LlamaStash in an LXC on Strix Halo is a fair setup and the admission gate shouldn't refuse models that fit there.

I ran the branch on my Strix Halo box, bare metal not LXC (4 GiB carve + 121.5 GiB GTT). Build, `make test` and clippy pass, LXC detection returns `false` there, and host metrics and the VRAM gauge (`0.9/116G`) are the same as on main. I dont have an LXC setup, so the container path itself isnt tested. The red `cargo audit` job is a `rustls` advisory (RUSTSEC-2026-0285), not from this PR.

These need changes before merge, details are inline:

- the `classify_amd_memory` change in `src/gpu/mod.rs` (it isn't LXC specific, and `used` can go above `total`)
- `amd_lxc_uma_uses_gtt_budget_instead_of_container_ram` passes without the LXC exception
- the rebase onto main needs `engine_pool_total_bytes` to use the same pool as the new free figure
- CPU-side launches in the container lose the RAM check

The smaller ones are inline too. Also, no docs mention the container exception, a short note next to the admission text in `docs/architecture.md` would help people who see different results in LXC and on bare metal.

Could you paste the four `mem_info_*` values from `/sys/class/drm/card*/device/` on the LXC host? That will show what the `gpu/mod.rs` change needs to handle.

deepu105:
**AI review follow-up.** I tested the open question from the review on my Strix Halo: does GPU memory from GTT count against a container memory limit? It doesnt, so the idea in this PR holds for offloaded models. The same test also confirms the CPU-side concern from the review.

Setup: `systemd-run --user -p MemoryMax=8G -p MemorySwapMax=0`, which sets the same cgroup v2 `memory.max` that an LXC `limits.memory` sets. Model `Qwen3.8-27B-UD-Q6_K.gguf` (20.5 GiB), page cache for the file dropped before each run. Kernel 7.2.4-arch1-2, ROCm 7.2.4, llama-server build 10905.

| Case | Result | cgroup `memory.peak` | GTT used |
|---|---|---|---|
| `-ngl 999` | loaded in 30s, 16 tokens in 3s, no OOM | 8.00 GiB (anon 1.43, file 6.55) | +20.28 GiB |
| `-ngl 0 -dev none` (mmap) | no OOM, 16 tokens took 225s | 8.00 GiB (file_mapped 6.58) | +0.02 GiB |
| `-ngl 0 -dev none --load-mode none` | OOM-killed during load (systemd `Result=oom-kill`) | 8.00 GiB (anon 6.96) | +0.02 GiB |

- GTT allocations aren't charged to `memory.max`. A 20 GiB fully offloaded model loads and serves under an 8 GiB limit.
- CPU-side memory is charged. With this PR, a CPU-only launch in an 8 GiB container is checked against the GTT pool and admitted, then it either gets OOM-killed or keeps re-reading model pages from disk.

So budgeting GTT for the offloaded part is correct, but `ram_free` still needs to limit the CPU part (inline comment on `amd_uma_lxc_pool_free`).

Not covered: this isn't a real LXC container, so the lxcfs `MemTotal`, the LXC detection and the admission code path weren't exercised. Partial offload wasn't tested either.

PR #84, opened 2026-09-12
fix(unit5): price a pool fraction against the pool, not against free

deepu105:
RV7 from the #79 review, now that #81 has landed and the reserve work is in `launch::admission`.

## The correction to my own note

I wrote this up as "over-refusal, so it fails safe" and that was wrong. There are two errors in the projection and they pull opposite ways:

| | direction |
|---|---|
| `free × fraction` instead of `pool × fraction` | **under**-refuses (free is always the smaller number) |
| gate adds the weights the fraction already covers | over-refuses |

The base error is the one that matters, and it is the dangerous direction. On a 121 GiB host with 52 GiB free, `gpu_memory_utilization: 0.9` was priced at **47 GiB**, fit inside the free reading, and got admitted. The launch takes about **109 GiB**. That is the configuration `docs/troubleshooting.md` already describes as having frozen this machine, and the gate was waving it straight through. The weights double-count only partly masked it.

Worth checking against the number already in the docs: the setup guide notes that even `0.15` reserved 15.1 GiB on a 0.5B. Old projection on that host: 7.8 GiB. New: 18.25 GiB. The measured figure sits between them, nearer the new one, and the old one was under it by half.

## The change

`projected_cache_bytes` could not be fixed without more inputs, so it takes a struct now:

```rust
pub struct DemandInputs {
  pub free_bytes: u64,        // what the demand is compared against
  pub pool_total_bytes: u64,  // what a fraction is a fraction OF
  pub weights_bytes: u64,     // what the gate is already counting
}
```

The pool total is a host fact rather than a backend one, so `engine_pool_total_bytes` derives it: system RAM when unified (that is the total these engines size against, and the measured `0.92 → ~106 GB of a 121 GB box` is the evidence), VRAM when discrete, `0` when the host cannot say. Pricing a fraction of *system RAM* on a discrete card would have been a second bug, which is why this is not just "pass the ram total".

Both engines call one shared `pool_fraction_beyond_weights`. The netting-off is not a special case either: the trait contract already said "bytes this launch will hold **beyond its weights**", and a knob covering weights *and* cache has to subtract them to honour that. I made the doc say so out loud, since it is the part that was silently violated.

With no pool total, free stands in as before. Understating is not good, but an engaged gate beats a skipped one, and that path is the unsampled window only.

## Host shapes, after review

The "free is the smaller number" premise is a **unified-host** one, and the review was right that I stated it too broadly. Per shape:

| host | before | after |
|---|---|---|
| unified | under-projected, admitted the freeze | correct |
| unified + GTT reported | under-projected | **over**-refuses every hand-set fraction (free is GTT-capped, projection is against full RAM) |
| discrete | over-refused (free sums VRAM + RAM) | under-tightens; the engine's own startup check refuses instead |

The GTT row is the one real finding and it is not fixed here. Aligning both sides on `uma_shared_total_bytes` is only correct if torch reports GTT as its device total on an ROCm APU; if it reports sysmem, the engine really attempts `fraction x full RAM` and pricing against GTT would understate exactly the freeze this function exists to project. Under-projection is the direction that takes the host down, so the conservative total stays until someone reads that number off real hardware. Both branches are in `TODO.md`, and the consequence plus the workaround (absolute cap, or `--force`) are in the setup docs.

## Worth a look

- **A byte cap is still checked first.** #81 writes a companion `gpu-memory-utilization` next to the auto byte cap, and the cap is the exact figure the launcher is handed while the companion is only a startup-check clearance — so the cap is what should be priced. `the_auto_cap_is_priced_by_its_bytes_not_its_companion_fraction` pins that ordering.

  *Corrected after review:* I first wrote that the fraction branch winning would "refuse every UMA launch". It would not. The companion is `min(weights + cap + engine overhead, free - context margin) / pool`, over the same free reading the gate compares against, so pricing it can never project past that reading — 16.7 GiB against 52 GiB free for this test's own numbers. It would misstate the demand by the clamp margin, in the over-stating direction. Wrong number, not a broken launch.
- **This makes the gate stricter for anyone who set a fraction by hand.** A `0.9` on a busy unified host now refuses where it used to launch. That is the point, and `--force` is the escape hatch, but it is a behaviour change a user can hit, so it is in the CHANGELOG in those terms.
- The fraction itself is unchanged — only what the gate *projects* from it. Nothing new reaches argv.

## Not in this PR

The admission ledger still prices a launch without the engine's own overhead, so two concurrent launches each under-reserve by ~6.5 GiB (Copilot found this on #81; it is in `TODO.md`). `DemandInputs` is the seam it needs and I deliberately did not widen it here on spec — that fix wants `unified` on the struct and a decision about discrete cards, where the 7 GiB was never measured.

## Verification

`make test` and `make lint` green, 47 test binaries. New coverage: the shared helper including the clamp and zero-pool paths, the pool-vs-VRAM selection per host kind, the byte-cap precedence, and one test per engine asserting the fraction no longer fits inside the free reading it used to be priced against. **Validated on GB10** by @bernardogv (vLLM 0.28, 121.69 GiB unified, beside a tenant leaving 51 GiB, Nemotron-3.5-Lightning-30B-A3B-NVFP4):

- `0.9` → refused before spawn at 110.0 GiB, exit 67. Under `--force`, vLLM's own check read `desired (0.9, 109.52 GiB)` — **the two sides agree to two decimals**, which is the confirmation the projection now models what the engine attempts.
- That also settles half the GTT question: torch takes MemTotal as its device total on NVIDIA coherent UMA, so `ram_total` is exact there and the ambiguity is AMD-only. Narrowed in the code comment and `TODO.md`.
- `0.2` → admitted at 24.8 GiB, served. Measured cost 26.8 GiB: ~2 GiB of CUDA context / API server sits outside the fraction's pool and `overhead_band_bytes` (512 MiB) does not reach it. Still a large improvement — the old arithmetic projected 10.2 GiB for that same launch — but it under-projects, so it is filed with the ledger follow-up that needs the same term.
- Also spotted, pre-existing and wider than this PR: a failing `--json` command prints the human `✗` line rather than a JSON object. `report` in `src/cli/mod.rs` is the single render site for every `CliExit` and never sees the flag, so it breaks the `--json` contract on every non-zero exit. Filed.

Strix Halo is still unrun, and it is the host shape where the GTT question actually bites.


src/launch/admission.rs
```diff
@@ -270,6 +270,68 @@ pub fn dir_weight_bytes(dir: &std::path::Path) -> u64 {
...
+/// `gpu_memory_utilization` / `mem_fraction_static` are shares of the device
+/// total, not of what is free and not of the model — so a projection needs the
+/// total, and which total depends on the host. On a unified host the device
+/// total *is* system RAM: the 0.92 default projected to ~106 GB of a 121 GB
+/// box, which is how these engines have frozen one. On a discrete host it is
+/// VRAM, and system RAM is irrelevant to the fraction.
+pub fn engine_pool_total_bytes(snap: &HostMetricsSnapshot) -> u64 {
+  let unified = snap.unified || snap.gpu_backend == HostMetricsSnapshot::BACKEND_APPLE_METAL;
+  if unified {
+    snap.ram_total_bytes
```
deepu105, line 313: On a GTT-carrying UMA host the two sides of the gate are denominated against different pools. `effective_free_bytes` budgets `min(ram_free, gtt_free)` because sysfs GTT is the budget authority, but this prices the fraction against `ram_total`. On a default-config APU, where GTT is roughly half of RAM (per that same doc comment), a hand-set `mem_fraction_static: 0.9` projects at ~0.9 x full RAM against a free reading that cannot exceed the GTT cap, so the gate refuses every hand-set fraction on that host even when the engine would fit inside GTT.

It fails safe, never under: `ram_total` is at least whatever the engine sees as its device total under either ROCm reading, so the projection never understates. But `uma_shared_total_bytes.unwrap_or(ram_total_bytes)` would put both sides on the same pool, and which of the two is *correct* depends on what torch reports as device total on an ROCm APU (GTT or sysmem). That is exactly the real-hardware check the PR body still owes, so worth deciding before that run rather than after.

src/launch/admission.rs
```diff
@@ -270,6 +270,68 @@ pub fn dir_weight_bytes(dir: &std::path::Path) -> u64 {
...
+  pub weights_bytes: u64,
+}
+
+/// An engine pool fraction priced for the admission gate: the bytes it will
+/// hold **beyond the weights** the gate already counts.
+///
+/// Two corrections live here, and they pull opposite ways, which is why the
+/// old `free * fraction` was not simply conservative:
+///
+/// - The fraction is of the pool, not of what is free. Free is always the
```
deepu105: Scope: this invariant only holds on unified hosts. On discrete, `effective_free_bytes` sums post-headroom VRAM free + system-RAM free, so free can exceed the VRAM pool the fraction is priced against: a 24 GiB card with 60 GiB of RAM idle gives free ~84 GiB against a pool of 24.

Consequences worth a doc line: old code multiplied by that summed free, so discrete fractions were over-refused; new code under-tightens (0.9 x 24 = 21.6 admitted against 84 free) and the refusal lands engine-side instead, since vLLM 0.28 compares `total x util` against its own free reading at `init_device`. Same outcome for safety, worse message for the user. Either way the sentence should say unified, not always.

src/launch/admission.rs
```diff
@@ -270,6 +270,68 @@ pub fn dir_weight_bytes(dir: &std::path::Path) -> u64 {
     .fold(0u64, u64::saturating_add)
 }
 
+/// The pool an engine's **fraction knob** is a fraction *of*, or `0` when the
+/// host cannot say.
+///
+/// `gpu_memory_utilization` / `mem_fraction_static` are shares of the device
+/// total, not of what is free and not of the model — so a projection needs the
+/// total, and which total depends on the host. On a unified host the device
+/// total *is* system RAM: the 0.92 default projected to ~106 GB of a 121 GB
```
deepu105: 0.92 x 121 GB is 111 GB. The ~106 figure comes from `docs/vllm-setup.md:169`, where it implies a ~115 GB total (106 / 0.92). As written the pair does not multiply out; either ~111 GB here, or keep 106 attached to its original box size.

src/backend/vllm/mod.rs
```diff
@@ -1115,19 +1123,52 @@ mod tests {
...
-    // An explicit byte cap is exact and wins.
+    // An explicit byte cap is exact, bounds the cache only, and wins.
     set(&mut p, "kv_cache_memory_bytes", "2G");
-    assert_eq!(b.projected_cache_bytes(&p, 100 * GB), Some(2 * GB));
+    assert_eq!(b.projected_cache_bytes(&p, &host), Some(2 * GB));
+  }
+
+  /// The companion utilization the guard writes beside an auto byte cap must
+  /// never be priced as a pool fraction — that would project most of the pool
+  /// for a launch the cap already bounds, and refuse every UMA launch.
```
deepu105: The precedence is worth pinning, but the stated failure mode does not hold. The guard's companion is `min(demand, effective_free - 3 GiB) / ram_total`, floored to two decimals, over the same `effective_free_bytes` the gate compares against. So even if the fraction branch won, pricing it gives at most `free - 3 GiB + band`, and the band is 0.5-1 GiB (`overhead_band_bytes`), which never crosses gate free. The fraction branch would misprice by the clamp margin, not refuse every UMA launch.

Worth correcting this comment and the "load-bearing / refuse every UMA launch" framing in the PR body: the ordering's real justification is that the cap is the exact figure and the companion is a startup-check clearance, not a demand. The test itself is fine to keep: it pins precedence, and the 0.13 fixture is exactly what `startup_utilization` writes for these numbers (min(1+8+7, 52-3) / 121 floored).

deepu105 (commented):
The core correction is right for the UMA case it targets: `frac x pool` netted against the weights the gate re-adds prices exactly what the engine takes, and the shared helper + `DemandInputs` is the right seam.

Findings: one pairing question on GTT hosts (over-refusal direction, fails safe), one comment that over-claims what the byte-cap ordering prevents, two doc nits. Nothing blocking.

src/launch/admission.rs
deepu105: Agreed on the mismatch, and I am keeping `ram_total` rather than switching — with the reasoning written into the function so the next person does not have to re-derive it.

The deciding factor is which way the error runs if the guess is wrong. If torch on an ROCm APU reports sysmem as its device total, the engine really does attempt `0.9 x full RAM`, and pricing against GTT would understate exactly the freeze this function exists to project. That is the direction that takes the host down, and it is the bug this PR is fixing, so I would rather over-refuse than risk re-introducing it on a guess.

So: conservative total stays, the over-refusal and its workaround (absolute `kv_cache_memory_bytes`, or `--force`) are now in `docs/vllm-setup.md`, and `TODO.md` carries both branches with the single measurement that settles it. You are right that it should be decided before the hardware run rather than after, and I think "decided" here means knowing which reading we are looking for, which is now written down.

src/launch/admission.rs
deepu105: Right, and the sentence was wrong as written. Fixed: the premise is scoped to unified hosts, and the discrete shape is spelled out in the same doc comment — free sums post-headroom VRAM free and system-RAM free, so it can exceed the VRAM pool, the old code over-refused, this one under-tightens and the refusal lands engine-side. I took your framing almost verbatim, including "same safety, worse message".

The PR body now carries a per-host-shape table too, since the summary had the same over-broad claim.

src/launch/admission.rs
deepu105: You are right, it does not multiply out. I dropped the product from my comment rather than restate a figure I cannot reconcile: 0.92 x 121 is 111, and the ~106 in `docs/vllm-setup.md:169` implies a ~115 GB total, so one of the two numbers in that original pair is against a different denominator than it looks.

I left the measured figure alone where it sits with its own box size, since it is someone else's measurement and I have no way to re-derive which denominator it used. Worth a look on the next GB10 run if the box size is recoverable.

src/backend/vllm/mod.rs
deepu105: Checked the arithmetic and you are right, my failure mode does not hold. For the test's own numbers the companion is `min(1 + 8 + 7, 52 - 3) / 121` floored, and pricing that gives a 16.7 GiB demand against 52 GiB free — no refusal, it would over-state by the clamp margin.

Corrected both places. The comment now gives the ordering's real justification (the cap is the exact figure, the companion is a startup-check clearance and not a demand) and says the mispricing is a wrong number rather than a broken launch. The PR body had the same "load-bearing / refuse every UMA launch" framing and now carries the correction explicitly. Test kept as-is.

deepu105:
@bernardogv i merged your other prs. could you plz help test this on nvidia :pray:

bernardogv:
Tested on a GB10 (DGX Spark, 121.69 GiB unified, vLLM 0.28) at `2b913c8`, beside a resident SGLang tenant leaving 51 GiB MemAvailable. Model: Nemotron-3.5-Lightning-30B-A3B-NVFP4, 20.1 GiB on disk.

- `--gpu-memory-utilization 0.9` → gate refuses at **110.0 GiB** (0.9 × 121.69 + band), exit 67, nothing spawned.
- Same with `--force` → vLLM's own check refuses: `Free memory on device cuda:0 (48.31/121.69 GiB) ... desired (0.9, 109.52 GiB)`. So the two sides agree to two decimals, and **torch reports the full unified pool (= MemTotal) as device total on NVIDIA coherent UMA** — `engine_pool_total_bytes = ram_total` is exact here. The GTT question stays AMD-only.
- `0.2` beside the same tenant → admitted at a 24.8 GiB projection, served. Measured launch cost 26.8 GiB at ready: the fraction covers weights + KV + graphs, but not the ~2 GiB of CUDA context / API server outside the pool, so the admit side is still slightly under. Same overhead term as the earlier byte-cap measurement; `overhead_band_bytes` at 0.5 GiB doesn't cover it.
- Minor, pre-existing: `start --json` on a gate refusal emits the `✗` human line rather than a JSON object.

Raw logs and the full write-up are in our casebook if you want them.

deepu105:
Thanks @bernardogv, thats the measurement that matters. `desired (0.9, 109.52 GiB)` against our 109.52 projection is the whole point of the PR confirmed from the other side — the gate and the engine now agree on what the fraction means.

The device-total finding is more useful than the refusal itself. I had left `engine_pool_total_bytes` on `ram_total` with a written-down bet that torch might report sysmem rather than GTT, and your run settles it for NVIDIA coherent UMA: MemTotal is the device total, so `ram_total` is exact there, not just conservative. Narrowed both the doc comment and the TODO to the AMD case, which is the only place the GTT carve makes the two sides disagree. Strix Halo is still the run that would close it.

Filed both gaps rather than widening this PR:

- The ~2 GiB out-of-pool footprint. Worth noting the fraction path improved from 10.2 GiB projected to 24.8 against a 26.8 actual, so this is the residual, not a regression — and it is the same term the ledger follow-up needs (concurrent launches under-reserve by the engine overhead). One fix should cover both, gated on a unified host since the number is unmeasured on a discrete card.
- The `--json` one is worse than it looks and I am glad you caught it. It is not specific to gate refusals: `report` in `src/cli/mod.rs` is the single render site for every `CliExit` and never sees the flag, so *any* non-zero exit gives an agent a `✗` string instead of the wrapped object `AGENTS.md` promises. Filed separately since it is a CLI-contract bug, not a memory one.

Pushed as b3e454e5 with the numbers in the plan record and the PR body. Yes please to the casebook link for the raw logs.

bernardogv:
Casebook page + raw logs: https://gist.github.com/bernardogv/c9601e43ee07daa5177ad5bf9ff4ece9

- `README.md` — the write-up (method, the three launches, the numbers above)
- `run-F1-0.9.log` — the gate refusal, exit 67
- `run-F3-0.2.log` / `vllm-F3-0.2.log` / `memlog-F3-0.2.tsv` — the served `0.2` run with the 1 Hz MemAvailable trace (26.76 GiB at ready vs 24.84 projected)
- `run-F3a-0.2-failed.log` / `vllm-F3a-0.2-failed.log` / `memlog-F3a-0.2-failed.tsv` — the first `0.2` attempt that died at graph capture on `max_num_seqs 256` (engine-side sizing, not the gate)

The `--force` run's vLLM `init_device` refusal (`desired (0.9, 109.52 GiB)`) is quoted verbatim in the page; I didn't keep a separate file for it. The page's link to the 2026-09-09 byte-cap measurement points into a private casebook — say if you want that one as well.

Read through `b3e454e` / `4d88359`: docs only, so the `2b913c8` validation stands.

deepu105:
Appreciate it @bernardogv 🙏

bernardogv:
> Appreciate it @bernardogv 🙏

Glad to help !

